PR #23, opened 2011-05-27
Support for the oracle_enhanced adapter

woodchuck:
Hi zdennis,

I've got all but 2 of the tests passing for oracle_enhanced.The failures are:

test: #import should import with a single insert(#import)
test: #supports_imports? should support import(#supports_imports?)

I disabled supports_import for oracle_enhanced because Oracle is rather more complicated in how it handles bulk inserts. Hopefully I'll have time to add benchmarking and experiment with some different approaches - parameter binding, pl/sql, sqlldr, etc.

I also had to change how you determine the current adapter because oracle_enhanced doesn't have a @config attribute. Hopefully connection.adapter_name.underscore also works for the other adapters.

Thanks!

-Steve


jkowens:
I'm closing this PR, but I would love to see an `activerecord-import-oracle_enhanced` gem created. See the README for some basic instructions to get started with that.

Once published feel free to add a wiki entry for additional adapters with a link to the gem/repo/etc.

PR #44, opened 2012-02-03
Sql server

ouranos:
Hi,

This branch adds basic support for SQL Server.

Some tests are failing but it's due to the fact that SQL Server doesn't accepts null insertion into primary keys.
ie:
`INSERT INTO [group] ([id],[order],[created_at],[updated_at]) VALUES (NULL,''superx'',''2012-02-03T17:32:29.796'',''2012-02-03T17:32:29.796'')`
should be:
`INSERT INTO [group] ([order],[created_at],[updated_at]) VALUES (''superx'',''2012-02-03T17:32:29.796'',''2012-02-03T17:32:29.796'')`

Tested with SQL Server 2008 and ruby 1.9.2

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,8 @@ group :test do
     gem "mysql2", "~> 0.2.4"
     gem "pg", "~> 0.9"
     gem "sqlite3-ruby", "~> 1.3.1"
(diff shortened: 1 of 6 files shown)


jkowens:
Thanks for your work. I'm closing this PR since this feature is continued on by #201.

PR #70, opened 2012-11-01
Sql server

btrepp:
Adds SQLServer support based of ouranos sql_server branch

This branch passes all tests. Though makes some sweeping (perhaps bad) assumptions. SQLServer does not allow NULL to be passed into the IDENTITY column. To get around this I have made instances of where the identity column is NULL ommit inserting the ID by removing it from the sql string.

This is making the assumption that the ID is the first column. It will probably cause issues if it is not the first column. But hey it passes all tests :)


jkowens:
Thanks for your work. I'm closing this PR since this feature is continued on by #201.

PR #85, opened 2013-02-28
Fix to support values with different encodings

cheerfulstoic:
Been running into this and it seems like when the values get quoted they can get turned into different kinds of encodings, which causes an error when they are combined.

Fix included with tests.  Let me know if you have any questions

### DIFF
--- a/test/import_test.rb
+++ b/test/import_test.rb
@@ -1,3 +1,5 @@
+# encoding: utf-8
+
 require File.expand_path('../test_helper', __FILE__)
 
 describe "#import" do
@@ -6,7 +8,7 @@ describe "#import" do
     assert_difference "Topic.count", +10 do
       result = Topic.import Build(3, :topics)
       assert result.num_inserts > 0
-    
+
       result = Topic.import Build(7, :topics)
       assert result.num_inserts > 0
     end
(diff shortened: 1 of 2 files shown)


jkowens:
@cheerfulstoic I know this is an old PR, but I'm trying to clean up the backlog. Do you think there still a need for this feature? If you could help me understand the purpose that would be great. Also the commit will need to be cleaned up a bit to resolve conflicts and it would be easier to review if the white space changes were eliminated. Thanks!

cheerfulstoic:
It's been a while, but I think it would still be a useful fix.  Github actually allows you to look at diffs without whitespace like so:

https://github.com/zdennis/activerecord-import/pull/85/files?w=1

Feel free to just copy those changes into `master` and see if things still work

test/import_test.rb
```diff
@@ -202,37 +204,49 @@
...
   end
-  
+
   context "with an array of columns and an array of values" do
     it "should import ids when specified" do
       Topic.import [:id, :author_name, :title], [[99, "Bob Jones", "Topic 99"]]
       assert_equal 99, Topic.last.id
     end
+
+    it "should import data with different encodings" do
```
jkowens: Should this spec fail without the change to `values_sql_for_columns_and_attributes` above? In my tests it always passes.

cheerfulstoic: Interesting... I would say that I think it should fail, but again it's been a while.  Maybe something changed lower down which is doing conversions for us now... ¯\_(ツ)_/¯

jkowens:
I'm going to close this for now, but if anyone runs into similar issues please feel free to reopen.

cheerfulstoic:
:+1:

PR #88, opened 2013-03-07
add delorean as a development dependency

pravi:
delorean is used for running tests


jkowens:
The timecop gem is currently being used for time travel, so I am closing for now.

Issue #22, opened 2011-05-09
#import adds timestamps to input array


jamiecobbett:
ActiveRecord::Base#import should not mutate (change) the array of data being imported, and there is a call to Object#dup to do this:
https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/import.rb#L197

Unfortunately, it doesn't seem to be working, as I experienced it adding timestamps to an array.

I haven't been able to figure out the cause, I suspect maybe ActiveRecord is being a little too helpful with its magic.

Sometimes calling dup on my dataset /before/ the call to #import avoided the problem, sometimes not; it seemed to differ between reboots of my Resque worker.

Reproduced on Ruby 1.8.7 & 1.9.2 with Rails 3.0.1 & 3.0.7.

I have written a failing test here:
https://github.com/jamiecobbett/activerecord-import/commit/7e45e23e386caa14f8605ee489263f798e0a7977

Test output:
      1) Failure:
    test: #import ActiveRecord timestamps when the timestamps columns are present should not alter the input dataset(#import ActiveRecord timestamps when the timestamps columns are present)
        [./test/import_test.rb:201:in `test: #import ActiveRecord timestamps when the timestamps columns are present should not alter the input dataset'
         /home/jamie/.rvm/gems/ruby-1.8.7-p334/gems/activesupport-3.0.1/lib/active_support/testing/setup_and_teardown.rb:67:in `__send__'
         /home/jamie/.rvm/gems/ruby-1.8.7-p334/gems/activesupport-3.0.1/lib/active_support/testing/setup_and_teardown.rb:67:in `run'
         /home/jamie/.rvm/gems/ruby-1.8.7-p334/gems/activesupport-3.0.1/lib/active_support/callbacks.rb:423:in `_run_setup_callbacks'
         /home/jamie/.rvm/gems/ruby-1.8.7-p334/gems/activesupport-3.0.1/lib/active_support/testing/setup_and_teardown.rb:65:in `run']:
    <[["LDAP", "Big Bird", "Del Rey"]]> expected but was
    <[["LDAP",
      "Big Bird",
      "Del Rey",
      Mon May 09 13:12:57 +0000 2011,
      Mon May 09 13:12:57 +0000 2011,
      Mon May 09 13:12:57 +0000 2011,
      Mon May 09 13:12:57 +0000 2011]]>.

    97 tests, 163 assertions, 1 failures, 0 errors
    rake aborted!
[...]

PR #92, opened 2013-03-23
No longer modify input array; deep clone it first

demands:
This pull request resolves issue #22.

If you call #import with an array of values, activerecord-import will currently modify that array with updated_at and created_at timestamps for each row. This change prevents the array you pass in from being modified.


jkowens:
This was included in #291 and can be closed.

PR #95, opened 2013-05-10
Optionally disable ON DUPLICATE KEY UPDATE

trusche:
For MySQL, it would be nice if the ON DUPLICATE KEY UPDATE option could be disabled completely, by passing `:on_duplicate_key_update => false`. 

Unfortunately I have no idea how to write a test for this small modification that uses your conventions; I have verified that it doesn't break any existing tests (for mysql2), and that it indeed drops the ON DUP... clause from the INSERT statement.

Many thanks for the great gem!

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -142,7 +142,7 @@ class ActiveRecord::Base
     #
     # == On Duplicate Key Update (MySQL only)
     #
-    # The :on_duplicate_key_update option can be either an Array or a Hash. 
+    # The :on_duplicate_key_update option can be either an Array or a Hash, or `false` to disable ON DUPLICATE KEY UPDATE completely. 
     # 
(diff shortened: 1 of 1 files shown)


nishanth980:
this should work 100 percent
AdminRole.import(columns, values, :on_duplicate_key_update => "updated_at = updated_at")

trusche:
Hi @zdennis -- is there anything I can do to improve this to the point where you'd consider merging it? I'd be happy to add tests if you can point me in the right direction. Thanks!

Empact:
This would be useful to me as well.

jkowens:
@zdennis this can be closed. It was implemented in 9b2e351437c7d9b496a47413f79e1f5b4b7eca41 and there is a [test](https://github.com/zdennis/activerecord-import/blob/master/test/support/shared_examples/on_duplicate_key_update.rb#L70).

PR #98, opened 2013-06-04
sqlserver and oracle adapters workaround

censo:
Hi. There was an error while trying to establish connection to mssql database. It was looking for mssqlserver adapter which is not supported yet. I have made some fix/workaround and everything works fine. Error doesn't occur any more.


jkowens:
This issue has since been fixed.

PR #104, opened 2013-07-03
return the ids from a bulk insert [postgres only]

robmathews:
A common request is to return the ids from the newly created object. This patch does that. It's postgres only, but at least it doesn't break for non-postgres databases - it just doesn't set the id. 

The api is unchanged. Instead, for every model that was saved, it sets the :id and marks the record clean, so that you can keep using it if you want.

Seems like one could pretty easily do a BFS on the changed associations and bulk-import the next level down in the tree in one sql statement as well .. at least for postgres.

Let me know if this is any use to you; works fine for us, and that's all I need.


alainmeier:
This is an excellent PR, will this be merged?

zdennis:
@alainmeier, it will become much more likely if it's become merge-able and stays green.

johnnaegle:
I've rebased the commits onto the latest master and got the tests working under mysql, postgresql and sqlite3.  The branch is here: https://github.com/GoodMeasuresLLC/activerecord-import/tree/set_ids_in_bulk_import.  I will open a PR from that branch when I think it is ready, but some early feedback would be helpful.

johnnaegle:
See #178

koheisg:
Why don't you close? it seems to be resolved for me.

zdennis:
Thanks for following up @koheisg, closing now.

PR #105, opened 2013-07-05
Settings ids in postgres

robmathews:
This PR fixes a problem where non-standard primary keys were causing errors.


johnnaegle:
see #178

jkowens:
@zdennis I believe this can be closed thanks to #178.

PR #118, opened 2013-11-24
validate using given models instead of re-initializing ones

giladas:
Hi, 
when using import with models and validations are enabled, the models are converted to arrays, re-initialized and then the validation process is taking place. It's not necessary to re-initialize the models since they were given as parameters, and it took more time than the actual db insert.


jkowens:
This has been implemented by #266.

PR #127, opened 2014-02-20
Add makara database adapter

jwg2s:
This makes activerecord-import compatible with the makara gem.  Thanks @cmthakur for the suggestion -- filename in your version was slightly off from newest version of makara.


zdennis:
Thanks @jwg2s ! The mysql2_makara has been merged into master as of c7424fd and will be available in the next release.

PR #135, opened 2014-03-26
Set "object_ids" during import

jzaleski:
If enabled (via "hydrate_object_ids"), set the "primary" and "foreign"
key values on the model objects passed into the "import" method. This
functionality was added in an attempt to make it easier to import graphs
of objects. This functionality only exists for the "PostgreSQL Adapter"


jkowens:
@zdennis I'm not entirely sure, but I think #178 replaced this issue?

PR #137, opened 2014-04-02
Allow for `:on_duplicate_key_update` to be turned off

MSeal:
There is currently no option to tell the MySQL importer to fail on duplicate entries. I added this support locally for rails 3 and decided to PR this feature back for the master fork. Now if you explicitly set `:on_duplicate_key_update => false` in the import options it won't allow for duplicate keys to update existing elements. This was important for an application where all the data being imported should be new rows and any duplicates indicates an error in the input data.

It would be nice if there was a way to create a new branch/release for rails 3 to get updates that the rails 4 master branch is receiving (so I can reference your master fork instead of the PR fork).

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -368,7 +368,7 @@ class ActiveRecord::Base
             array_of_attributes.each { |arr| arr << value }
           end
 
-          if supports_on_duplicate_key_update?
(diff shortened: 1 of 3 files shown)


undergroundwebdesigns:
Can this be merged? I'd like to use this functionality, and am currently having to hack around it.

Thanks!

zdennis:
Thanks @MSeal. Sorry for the long wait on merging this in.

Issue #140, opened 2014-04-24
{ClassName}.import causes datetime fields to be validated incorrectly due to {field_name}_before_type_cast


trandoanhung1991:
Currently I'm using AR-import 0.5.0 in my project and I encountered this problem.

The class I'm using to import has a datetime field. When I call import, due to the way the attributes are taken out using _before_type_cast, the UTC time is taken out. However, when assigning the attributes, line 257, ```model.send("#{k}=", v)``` is used, which causes the UTC time to be parsed once again from the timezone into UTC.

This messes up a validation where I check the datetime field for duplication, since the timestamp is potentially non-existent.

A possible fix is to assign the attributes directly by using ```model[k] = v```, though I'm not sure if it'll break other things.

jkowens:
@zdennis this can be closed because #259 has been merged.

zdennis:
:thumbsup:

PR #141, opened 2014-04-25
Assign attributes directly during import validation

trandoanhung1991:
To prevent parsing UTC time into local datetime


README.markdown
```diff
@@ -1,76 +1 @@
...
-
-* Blythe Dunham
-* Gabe da Silveira
-* Henry Work
-* James Herdman
-* Marcus Crafter
-* Thibaud Guillaume-Gentil
-* Mark Van Holstyn 
-* Victor Costan
+Tioeuaeuesting webhook to Jenkins
```
jkowens: I don't think you meant to commit this?

jkowens:
@trandoanhung1991 If you remove your change to README.markdown and squash this into a single commit that would probably help move this along. I think there's been another issue discovered that this fix could resolve. See #235. 

I would also suggest referencing the related issue #140 that you opened in the description. :smile:

trandoanhung1991:
For some reason I can't find this repo anywhere. I'll close this PR and create a new one, is that ok?

trandoanhung1991:
Just in case, the new PR is #237

PR #143, opened 2014-05-22
Prequote timestamps for a speed boost

chewi:
I did some profiling and found that a lot of time is spent quoting the timestamps for every single row, even though the result is the same each time. We can prevent this from happening by quoting them beforehand instead. This gave me about a 5% speed boost.


chewi:
Rebased for your convenience. Travis failed but I don't believe that was my fault.

zdennis:
Thanks @chewi, I don't think it's related to your change either. I'll look at this weekend.

PR #144, opened 2014-06-12
Handle primary_key array (compositive_foreign_key)

### LINKED ISSUE #111: Support composite_primary_keys gem
If the first import method is used, a `SystemStackError (stack level too deep)` occurs
**NOTE**: Composite Primary Key is `id` + `book_id`

**code block**

``` ruby
books = []
10.times do |i|
  books << Book.new({:book_id => i, :title => "title #{i}", :desc => "desc #{i}"})
end
Book.import books
```

**error log**

```
SystemStackError (stack level too deep):
  actionpack (4.0.0) lib/action_dispatch/middleware/reloader.rb:70
```
### Successful code block

The following code works though (second import method):

``` ruby
books = []
books_keys = [ :book_id, :title, :desc ]
book_id = 1001;
10.times do |i|
  books << [ book_id + i, "title #{i}", "desc #{i}" ]
  # logger.info("Added book #{i}")
end
logger.info books
Book.import books_keys, books
```

chuckd:
Ripped from an old Lighthouse ticket: http://zdennis.lighthouseapp.com/projects/14379/tickets/30-patch-to-support-importing-tables-with-composite-primary-keys

I'll do some further testing and try to add a test case for this when I get a chance. Putting it out there for now for discussion.

This may fix #111.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -205,9 +205,10 @@ class ActiveRecord::Base
       # Force the primary key col into the insert if it's not
       # on the list and we are using a sequence and stuff a nil
       # value for it into each row so the sequencer will fire later
-      if !column_names.include?(primary_key) && connection.prefetch_primary_key? && sequence_name
-         column_names << primary_key
-         array_of_attributes.each { |a| a << nil }
+      if ![primary_key].flatten.*.to_sym.to_set.subset?(column_names.*.to_sym.to_set) &&
+          sequence_name && connection.prefetch_primary_key?
+        column_names << primary_key
+        array_of_attributes.each { |a| a << nil }
       end
 
       # record timestamps unless disabled in ActiveRecord::Base


jufemaiz:
Any thoughts on implementation?

lib/activerecord-import/import.rb
```diff
@@ -205,9 +205,10 @@ def import( *args )
       # Force the primary key col into the insert if it's not
       # on the list and we are using a sequence and stuff a nil
       # value for it into each row so the sequencer will fire later
-      if !column_names.include?(primary_key) && connection.prefetch_primary_key? && sequence_name
-         column_names << primary_key
-         array_of_attributes.each { |a| a << nil }
+      if ![primary_key].flatten.*.to_sym.to_set.subset?(column_names.*.to_sym.to_set) &&
```
jkowens: Does that line of code work? I think something like this is a little easier to grok:

```
symbolized_column_names = Array(column_names).map(&:to_sym)
symbolized_primary_key = Array(primary_key).map(&:to_sym)

if !symbolized_primary_key.to_set.subset?(symbolized_column_names.to_set) && ..
```

jkowens:
Just to help others that come along, this functionality depends on the `composite_primary_keys` gem. ActiveRecord does not provide support for composite primary keys. I would definitely be interested in seeing some tests added for this. 

For databases that support recursive import, a test to see if these would work together would also be nice.

jkowens:
I'm closing for now due to inactivity. If someone decides to pick this up, tests are needed to confirm compatibility with the `recursive`, `on_duplicate_key_update` and `on_duplicate_key_ignore` options.

Issue #150, opened 2014-07-08
ActiveRecord::Store values are not encoded


iamWh1sp3r:
I have a model with `store :my_store, accessors: [:my_attr], coder: JSON`.
The value gets saved as "--- !ruby/hash:ActiveSupport::HashWithIndifferentAccess".

sgil-carecloud:
having same problem

DWC01:
Hey all, 

Having same issue here, anyone find a fix to this?

ben-av:
I encountered the same issue.
Looks like using accessors is not supported in AR-Import.

Managed to get around it by skipping the accessors and providing a hash.
```
Model.new(my_store: {my_attr: 'test'}) # works
Model.new(my_attr: 'test') # attribute will not be saved
```

jkowens:
Store accessors should now be supported.

PR #151, opened 2014-07-19
Add Hstore support

ghiculescu:
Adds support for Rails 4 Postgres Hstores.

With Hstores, calling `#{name}_before_type_cast` always returns an empty string (I haven't dug in enough to find out why), but calling `name` directly returns a hash, which is then transformed into an insert statement correctly in the `values_sql_for_columns_and_attributes` method.

If the user isn't using Postgres or there are no Hstore columns, they won't even notice the difference.

I wasn't able to get the tests running (the MySQL gem wouldn't install and I don't really care to work out why) but I have this running successfully in my project. That said it is probably worth adding tests if anyone else is able to do so.


leonardvandriel:
Hi, I can confirm this issue. This fix works from me too, thanks. Any plans on merging?

brodock:
:+1:

jkowens:
Hstore columns are currently supported, it's just that the accessors are not. When importing, the entire hash value for the store must be set:

```
Model.new(my_store: {my_attr: 'test'}) # works
Model.new(my_attr: 'test') # attribute will not be saved
```
See #150.

Besides `hstore` there is `jsonb` as well as `text` that can contain serialized data. Is there a way to make this a more complete solution? Are there performance tradeoffs that need to be considered? Setting store values using a hash doesn't seem like it would necessarily be that much more trouble.

mparodi:
The same happens for JSONB columns! I would change one of your lines of the suggested marge like this:

```
if columns_hash.has_key?(name) && [:hstore, :jsonb].include?(columns_hash[name].type)
```

I hope to see this on the next version of activerecord-import!

jkowens:
Replaced by #322.

PR #153, opened 2014-08-11
Sum array explicitly to avoid ActiveSupport bug

ihat:
In activesupport (4.1.4), there's a regression in Array#sum
```
[1] pry(main)> ['a', 'b'].sum { |x| x.bytesize }
=> "ab"
[2] pry(main)> ['a', 'b'].map { |x| x.bytesize }.sum
=> 2
```

This commit is a workaround.


zdennis:
Closing as I can't reproduce.

PR #164, opened 2014-11-10
Rails 4.1 enum supported

urokuta:
I've implemented conversion code from enum value to integer.


zdennis:
Thanks @urokuta and @wadako111 . I believe this is resolved as part of this commit: a77b192300a1aaf0fe32a10640ad3fa90caa8aa5

Issue #159, opened 2014-09-17
Import has_many scope produces orphans


timscott:
I expect this to work:

``` ruby
class Foo
  has_many :bars
  def import_data(data)
    to_insert = data.map { |x| Bar.new a: x[:a], b: x[:b] }
    bars.import to_insert
  end
end
```

In the resulting insert statement the value for `foo_id` is always `NULL` resulting in all orphans.  Is there a bug, or am I doing it wrong?

PR #172, opened 2014-12-22
Adding the ability to use #import thru a has many association.

zdennis:
Closes Issue #159 .


zdennis:
This was included in the activerecord-import 0.7.0 gem release. 0.8.0 was just released now so it's included in that as well.

Issue #139, opened 2014-04-23
Rails 4.1 enums are not working


thomas-mcdonald:
From experimenting with my importer today it appears that enum fields are being inserted as `0` regardless of the string/integer value assigned to them.

zamirkhan:
That fix works for me - thanks!

iamWh1sp3r:
This is a must have

sebastianvera:
:+1:

zdennis:
I'm all for this, we need an associated test though, not just the implementation. If someone can update #164 with a test, this will get pulled in much faster.

dbenjamin57:
+1, we wasted half a day until googling to find there's an issue with enums. Makes this gem un-useable

sinelaw:
+1 isn't this bug somewhat...critical?

arashm:
+1
Made a lot of changes on my code to use this, and then this bug...

vmendoza:
please fix this.

zdennis:
@arashm deserves all the credit for fixing this. A 0.10.0 gem has been pushed to ruby gems incorporating his fix.

PR #173, opened 2014-12-27
Use read_attribute_before_type_cast instead of #{name}_before_type_cast and allow enums to work

jthibeaux:
The Rails 4.1 enum library overrides the #{name}_before_type_cast methods so that they return strings that cannot be inserted into the integer columns.  By making this change, the override is avoided and the code reads better anyway.

Addresses issue: https://github.com/zdennis/activerecord-import/issues/139


zdennis:
Thanks @jthibeaux for reporting this way back when. I believe this is resolved now so I'm going to close out this issue.

PR #177, opened 2015-02-09
Wrong usage of ActiveSupport.on_load

amatsuda:
activerecord-import is using `ActiveSupport.on_load` in a wrong way.

`ActiveSupport.on_load` is not just a way to add an additional block of code into a Ruby method.
It's a mechanism for adding load hooks which is needed to implement lazy loading in ActiveSupport.

And, in this particular "after_established" case, we don't actually need any special mechanism. Let's just simply call the logic inside the AMC.

This patch is put on top of another PR #136 which I sent last year, because these two PRs are tackling the same sort of problem, and editing the same file. So please just merge this one, then I'll close another one.

### DIFF
--- a/lib/activerecord-import.rb
+++ b/lib/activerecord-import.rb
@@ -1,10 +1,12 @@
-class ActiveRecord::Base
-  class << self
-    def establish_connection_with_activerecord_import(*args)
-      establish_connection_without_activerecord_import(*args)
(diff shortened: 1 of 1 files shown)


johnnaegle:
Failing tests might be fixed by #180

zdennis:
Merged in #227

amatsuda:
@zdennis :sparkles:

PR #179, opened 2015-02-21
Relax restrictions on mysql for AR 3.1 and 3.2

johnnaegle:
Allow any version of mysql > 2.8.1 instead of just 2.8.1.  2.8.2 was yanked and 2.8.1 does not seem to build on OS X 10.10


johnnaegle:
Failing tests should be fixed by #180

zdennis:
Thanks @johnnaegle

zdennis:
This has been included in the activerecord-import 0.8.0 gem release.

PR #183, opened 2015-03-10
Get Benchmarks to Run

johnnaegle:
Get benchmarks to run, add defaults to the option parser,  benchmark.rb borrowed a lot from test_helper.rb


zdennis:
Thanks for updating the benchmarks to run and the README to indicate how to run them!

PR #184, opened 2015-04-15
Improve support of STI models import

relev:
Thanks for the gem.

1st commit:

AR.scoped_attributes has their name as string, so when appended to column_names must be symbolized (otherwise includes?() won't detect duplicates).

2nd commit:

AR adds to scoped attributes
{
type: 'STIModelClass'
"type" => ['STIModelClass','STIModelParentClass']
}

The :type is ok for us,
the second one is used for building STI search requests, but cannot be used for insert.
So I just ignore it.


zdennis:
Thanks @relev, I updated your commit(s), kept your authorship, and merged in as part of  #225.

PR #185, opened 2015-04-24
Fix global constant lookup

antoshalee:
By default const_defined? looks for constant in all ancestors of the module including Object. 
In my case I included another gem in my project with global namespace "Import". After that activerecord-import stopped to require own code because ActiveRecord.const_defined?(:Import) returns true.


zdennis:
Thanks @antoshalee. I just pushed a 0.9.0 release which includes this.

PR #193, opened 2015-06-18
Fix a typing issue to dedupe column_names on import

zenspider:
I have a client that is using AR-import on an STI table and is upgrading to rails 4. This code breaks on upgrade because `scope_attributes` returns both `:type` and `"type"`. Since de-duping is happening via an `include?` check against a symbol, the inserted value should be cast to a symbol as well.


zdennis:
Thanks @zenspider. I just pushed a 0.9.0 release which includes this.

zenspider:
Thanks!

PR #195, opened 2015-06-23
Switch sqlite3-ruby to sqlite3 gem

fenec:


zdennis:
Thanks @fenec! Any reason for 1.3.3 specifically (I see 1.3.10 is the latest for sqlite3)?

fenec:
@zdennis it's only because I saw that sqlite3-ruby depends on sqlite3 >= 1.3.3. so I thought it would be safe to use this oldest possible major version with pessimistic versioning. I will bump it to the latest 1.3.10 then.

zdennis:
Merging this in now @fenec. Thank you!

PR #197, opened 2015-07-07
Fix Rails 4.2 Enum Support, fix for #139

arashm:
Sorry, I made a mess on my fork which made me reset and recreate the pull request.
I also added two tests for importing enum fields with both integer value and label (Rails > 4.1).


zdennis:
Thank you @arashm ! This is very much appreciated!

PR #198, opened 2015-08-02
Fix typos inside comments for return details

snkashis:


zdennis:
Whoops! I feel like face palming right now. Thank you for spotting them @snkashis and fixing them.

PR #75, opened 2013-01-31
Adds JDBC Support for MS-SQL

Altonymous:

PR #201, opened 2015-09-01
Support for SQL Server

keeguon:
A long time ago this PR was made: https://github.com/zdennis/activerecord-import/pull/75, I decided to improve on it (especially the part where it should check where the `id` is and if the matching value is  `NULL` or not).

However, I didn't took the time to write the tests and I haven't added the SQL Server gems when using MRI since they rely on a lib that is only available on Windows and would make the gem unusable on other OS.

I'm open to any reviews on this in order to get it merged ASAP.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ platforms :jruby do
   gem "jdbc-postgres"
   gem "activerecord-jdbcmysql-adapter"
   gem "activerecord-jdbcpostgresql-adapter"
+  gem "activerecord-jdbcmssql-adapter"
 end
 
 # Support libs
--- a/lib/activerecord-import/active_record/adapters/sqlserver_adapter.rb
+++ b/lib/activerecord-import/active_record/adapters/sqlserver_adapter.rb
@@ -1,7 +1,7 @@
 if (RUBY_PLATFORM =~ /java/).nil?
   require "active_record/connection_adapters/sqlserver_adapter"
 else
-  require "active_record/connection_adapters/jdbcmssql_adapter"
+  require "active_record/connection_adapters/mssql_adapter"
 end
 require "activerecord-import/adapters/sqlserver_adapter"
 
--- a/lib/activerecord-import/active_record/adapters/sqlserver_adapter.rb
+++ b/lib/activerecord-import/active_record/adapters/sqlserver_adapter.rb
@@ -1,0 +1,16 @@
+if (RUBY_PLATFORM =~ /java/).nil?
+  require "active_record/connection_adapters/sqlserver_adapter"
+else
+  require "active_record/connection_adapters/jdbcmssql_adapter"
+end
+require "activerecord-import/adapters/sqlserver_adapter"
+
+if (RUBY_PLATFORM =~ /java/).nil?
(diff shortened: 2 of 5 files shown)


zdennis:
Hi @Keeguon, I apologize for the delayed response I would love to merge this in. I haven't been on Windows in quite some time. Would you be able to provide any guidance on how I can up a dead simple environment for running this?

Thank you for the time and effort you spent into getting this going.

Keeguon:
Hi @zdennis, I could probably provide you with a VirtualBox VM (which is what I'm using to run tests on this). It won't be before Monday though and it could be large payload but if you're okay with it I'll send you some kind of OVA package or something.

thoughtafter:
I just tried this patch set on a project I'm working on. It seems to work great. I'm using MRI on Linux with activerecord-sqlserver-adapter. I would like to see this merged and if there is anything I can do to help, let me know.

zdennis:
@thoughtafter, I just need a way to run this over time so we can be sure to not break the compatibility with new activerecord-import updates. If we can run it on travisci or circleci I would be very interested in doing that. Worst case scenario it'd be nice to have a Amazon EC2 VM (or similar) that could be fired up to run the tests against. @Keeguon was working on something but life got busy.

Or if anyone has any other thoughts I'm all ears.

Keeguon:
Hey!

Yeah sorry. Will upload an OVA for you at the office today.

Keeguon:
@zdennis didn't forget you but yesterday was a little busy. I created a VirtualBox Appliance with Windows 10 and SQL Server 2014 which expires in 90 days. It is available here: 
https://mega.nz/#!GNcyVBSA (~10GB). The SQL Server is configured to be accessed w/ the Windows user (no independant login/password). If you run into some issues let me know.

jkowens:
@zdennis I think this might be a good fit for an external gem. Then in the README or Wiki for activerecord-import there could be a section that points to external adapters for other databases. It looks like [Appveyor](https://www.appveyor.com) could be used for testing builds.

zdennis:
Thanks @Keeguon for the VBA. I tried to download it but I'm asked for a decryption key before I can do that. Can you share that or am I doing something wrong?

I tend to agree with @jkowens thoughts that this would make for a great external gem with someone who can maintain it in a Windows environment. But, with MS announcing [SQLServer coming to Linux ](https://www.microsoft.com/en-us/server-cloud/sql-server-on-linux.aspx) it may be inevitable that this gets pulled back in. But may we go external gem first then pull in when we can setup SQL Server on Linux.

Thoughts?

jkowens:
@keeguon I would love to see this become the `activerecord-import-sqlserver` gem. I think there are many that could benefit from this functionality, but for now it would be hard for us to support along with the core activerecord adapters. Instead of this just sitting here in the queue, would you still be open to taking that on?

I think there may be some updates you might be able to make use of such as the ActiveRecord::Import::ValueSetsRecordsParser. SQLite has a 500 record limit when it comes to multiple values insert, so you may want to compare how that was implemented since SQL Server has a 1000 value limit.

jkowens:
I'm going to go ahead and close this PR, but again I think it would be great if this was packaged up into a separate gem.

PR #204, opened 2015-10-17
Add postgresql_makara_adapter

ptrln:
This adds support for MakaraPostgreSQLAdapter.

### DIFF
--- a/lib/activerecord-import/active_record/adapters/postgresql_makara_adapter.rb
+++ b/lib/activerecord-import/active_record/adapters/postgresql_makara_adapter.rb
@@ -1,0 +1,5 @@
+require "activerecord-import/adapters/postgresql_adapter"
+
+class ActiveRecord::ConnectionAdapters::MakaraPostgreSQLAdapter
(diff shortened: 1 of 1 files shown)


zdennis:
@ptrln, where can I find the `postgresql_makara` adapter? I don't see any makara gems doing a `gem search -r makara`

ptrln:
We're using the `makara` gem with PG, which includes the adapter here: https://github.com/taskrabbit/makara/blob/master/lib/active_record/connection_adapters/makara_postgresql_adapter.rb

zdennis:
Just noting that when we pull this in we're going to update this to follow conventions in https://github.com/zdennis/activerecord-import/pull/274 and to also make sure  https://github.com/zdennis/activerecord-import/blob/master/Rakefile#L16 and https://github.com/zdennis/activerecord-import/blob/master/.travis.yml#L40 are updated so the test suite runs against it.

zdennis:
Thanks @ptrln ! I pulled merged this into master in 8ee0133

PR #205, opened 2015-10-20
Separate associations in example schema

barendt:
This just adds a little formatting to separate the two associations in the hypothetical schema to make it a little easier to read.


zdennis:
Thank you @barendt

PR #206, opened 2015-10-28
Set new_record to false after set model ids

fpenapita:
This is needed to avoid that rails try to save a new object of this
model if later you use one of the imported objects (with the id set) in
an association with other new model that is being saved in database.


zdennis:
Thanks @fpenapita !

zdennis:

PR #209, opened 2015-11-06
quick and dirty support for postgres ON CONFLICT DO

ptolts:


zdennis:
@ptolts thanks for your work here. I think PR #218 covers this so I'm going to close this, but if you think that may not be the case don't hesitate to let me know. Thanks,

PR #213, opened 2015-12-09
Support MySQL strict mode, and use for tests

MaxGabriel:
Currently the MySQL test suite explicitly disables MySQL strict mode. The test suite passes in strict mode with changes to only one test though, and strict mode is now the default [as of Rails 4](http://guides.rubyonrails.org/4_0_release_notes.html), so it seems preferable to test in strict mode.


zdennis:
Thanks @MaxGabriel !

PR #217, opened 2016-01-11
postgresql adapter clears cache on insert_many

psobocinski:
Commit ee7411acaf32db233edb611e3d0ed885a2b589f5 defined an `insert_many` method in `ActiveRecord::Import::PostgreSQLAdapter`. As this method uses `select_values`, the ActiveRecord query cache is not invalidated after the query is executed.

This pull request invalidates the cache by calling `ActiveRecord::Base.connection.query_cache.clear` immediately after `select_values` is called.


zdennis:
Thank you @psobocinski  for this fix!

zdennis:

PR #218, opened 2016-01-12
Add :on_duplicate_key_update support for Postgres

### LINKED ISSUE #187: Support INSERT ... ON CONFLICT DO NOTHING/UPDATE (also known as UPSERT) for PostgreSQL
The feature exists already for MySQL databases and can be used with the :on_duplicate_key_update parameter. Upcoming versions of PostgreSQL will feature a similar functionality: 

http://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=168d5805e4c08bed7b95d351bf097cff7c07dd65

It would be cool to see this feature in activerecord-import, also for PostgreSQL databases.

jkowens:
Postgres 9.5 added UPSERT support. Resolves #187.

Provides the following syntax:

```
Country.import [country], validate: false, on_duplicate_key_update: { conflict_target: :country, columns: [:updated_at] }
```

`:conflict_target` takes either a single value or an array of column names that make up the unique constraint and `:columns` could be either an array or hash similar to the existing MySQL functionality.

### DIFF
--- a/lib/activerecord-import/adapters/postgresql_adapter.rb
+++ b/lib/activerecord-import/adapters/postgresql_adapter.rb
@@ -1,5 +1,8 @@
 module ActiveRecord::Import::PostgreSQLAdapter
   include ActiveRecord::Import::ImportSupport
+  include ActiveRecord::Import::OnDuplicateKeyUpdateSupport
+
+  MIN_VERSION_FOR_UPSERT = 90500
 
   def insert_many( sql, values, *args ) # :nodoc:
     number_of_inserts = 1
@@ -24,10 +27,111 @@ module ActiveRecord::Import::PostgreSQLAdapter
 
   def post_sql_statements( table_name, options ) # :nodoc:
     unless options[:primary_key].blank?
-      super(table_name, options) << (" RETURNING #{options[:primary_key]}")
+      super(table_name, options) << ("RETURNING #{options[:primary_key]}")
     else
       super(table_name, options)
     end
+  end
+
+  # Add a column to be updated on duplicate key update
+  def add_column_for_on_duplicate_key_update( column, options={} ) # :nodoc:
+    arg = options[:on_duplicate_key_update]
+    if arg.is_a?( Hash )
+      columns = arg.fetch( :columns ) { arg[:columns] = [] }
+      case columns
+      when Array then columns << column.to_sym unless columns.include?( column.to_sym )
+      when Hash then columns[column.to_sym] = column.to_sym
+      end
+    elsif arg.is_a?( Array )
+      arg << column.to_sym unless arg.include?( column.to_sym )
+    end
+  end
+
+  # Returns a generated ON CONFLICT DO NOTHING statement given the passed
+  # in +args+.
+  def sql_for_on_duplicate_key_ignore( table_name, *args ) # :nodoc:
+    arg = args.first
+    if arg.is_a?( Hash )
+      conflict_target = sql_for_conflict_target( arg )
+    end
+    " ON CONFLICT #{conflict_target}DO NOTHING"
+  end
+
+  # Returns a generated ON CONFLICT DO UPDATE statement given the passed
+  # in +args+.
+  def sql_for_on_duplicate_key_update( table_name, *args ) # :nodoc:
+    arg = args.first
+    if arg.is_a?( Array ) || arg.is_a?( String )
+      arg = { :columns => arg }
+    end
+    return unless arg.is_a?( Hash )
+
+    sql = " ON CONFLICT "
+    conflict_target = sql_for_conflict_target( arg )
+
+    columns = arg.fetch( :columns, [] )
+    if columns.respond_to?( :empty? ) && columns.empty?
+      return sql << "#{conflict_target}DO NOTHING"
+    end
+
+    conflict_target ||= sql_for_default_conflict_target( table_name )
+    unless conflict_target
+      raise ArgumentError, 'Expected :conflict_target or :constraint_name to be specified'
+    end
+
+    sql << "#{conflict_target}DO UPDATE SET "
+    if columns.is_a?( Array )
+      sql << sql_for_on_duplicate_key_update_as_array( table_name, columns )
+    elsif columns.is_a?( Hash )
+      sql << sql_for_on_duplicate_key_update_as_hash( table_name, columns )
+    elsif columns.is_a?( String )
+      sql << columns
+    else
+      raise ArgumentError, 'Expected :columns to be an Array or Hash'
+    end
+    sql
+  end
+
+  def sql_for_on_duplicate_key_update_as_array( table_name, arr )  # :nodoc:
+    results = arr.map do |column|
+      qc = quote_column_name( column )
+      "#{qc}=EXCLUDED.#{qc}"
+    end
+    results.join( ',' )
+  end
+
+  def sql_for_on_duplicate_key_update_as_hash( table_name, hsh ) # :nodoc:
+    results = hsh.map do |column1, column2|
+      qc1 = quote_column_name( column1 )
+      qc2 = quote_column_name( column2 )
+      "#{qc1}=EXCLUDED.#{qc2}"
+    end
+    results.join( ',' )
+  end
+
+  def sql_for_conflict_target( args={} )
+    if constraint_name = args[:constraint_name]
+      "ON CONSTRAINT #{constraint_name} "
+    elsif conflict_target = args[:conflict_target]
+      '(' << Array( conflict_target ).join( ', ' ) << ') '
+    end
+  end
+
+  def sql_for_default_conflict_target( table_name )
+    "(#{primary_key( table_name )}) "
+  end
+
+  # Return true if the statement is a duplicate key record error
+  def duplicate_key_update_error?(exception)# :nodoc:
+    exception.is_a?(ActiveRecord::StatementInvalid) && exception.to_s.include?('duplicate key')
+  end
+
+  def supports_on_duplicate_key_update?(current_version=self.postgresql_version)
+    current_version >= MIN_VERSION_FOR_UPSERT
+  end
+
+  def supports_on_duplicate_key_ignore?(current_version=self.postgresql_version)
+    supports_on_duplicate_key_update?(current_version)
   end
 
   def support_setting_primary_key_of_imported_objects?
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -116,7 +116,7 @@ class ActiveRecord::Base
     # supports on duplicate key update functionality, otherwise
     # returns false.
     def supports_on_duplicate_key_update?
-      connection.respond_to?(:supports_on_duplicate_key_update?) && connection.supports_on_duplicate_key_update?
+      connection.supports_on_duplicate_key_update?
     end
 
     # returns true if the current database connection adapter
@@ -165,15 +165,19 @@ class ActiveRecord::Base
     # below for what +options+ are available.
     #
     # == Options
-    # * +validate+ - true|false, tells import whether or not to use \
+    # * +validate+ - true|false, tells import whether or not to use
     #    ActiveRecord validations. Validations are enforced by default.
-    # * +on_duplicate_key_update+ - an Array or Hash, tells import to \
-    #    use MySQL's ON DUPLICATE KEY UPDATE ability. See On Duplicate\
-    #    Key Update below.
+    # * +ignore+ - true|false, tells import to use MySQL's INSERT IGNORE
+    #    to discard records that contain duplicate keys.
+    # * +on_duplicate_key_ignore+ - true|false, tells import to use
+    #    Postgres 9.5+ ON CONFLICT DO NOTHING.
+    # * +on_duplicate_key_update+ - an Array or Hash, tells import to
+    #    use MySQL's ON DUPLICATE KEY UPDATE or Postgres 9.5+ ON CONFLICT
+    #    DO UPDATE ability. See On Duplicate Key Update below.
     # * +synchronize+ - an array of ActiveRecord instances for the model
     #   that you are currently importing data into. This synchronizes
     #   existing model instances in memory with updates from the import.
-    # * +timestamps+ - true|false, tells import to not add timestamps \
+    # * +timestamps+ - true|false, tells import to not add timestamps
     #   (if false) even if record timestamps is disabled in ActiveRecord::Base
     # * +recursive - true|false, tells import to import all autosave association
     #   if the adapter supports setting the primary keys of the newly imported
@@ -211,7 +215,7 @@ class ActiveRecord::Base
     #  BlogPost.import posts, :synchronize => posts, :synchronize_keys => [:title]
     #  puts posts.first.persisted? # => true
     #
-    # == On Duplicate Key Update (MySQL only)
+    # == On Duplicate Key Update (MySQL)
     #
     # The :on_duplicate_key_update option can be either an Array or a Hash.
     #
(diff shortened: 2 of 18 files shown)


zdennis:
OMG @jkowens, I :heart: you right now.

jkowens:
Haha! Let me know if you find any issues or have any suggestions for this PR, and I'll do what I can. Thanks :beers:

triptec:
Any progress here?

jkowens:
I've rebased master and made some tweaks to the PR. If anyone else has a chance to review and test that would probably help @zdennis out a bit :wink:

lib/activerecord-import/adapters/postgresql_adapter.rb
```diff
@@ -30,6 +31,67 @@ def post_sql_statements( table_name, options ) # :nodoc:
...
+    end
+  end
+
+  # Returns a generated ON CONFLICT DO UPDATE statement given the passed
+  # in +options+.
+  def sql_for_on_duplicate_key_update( table_name, options={} ) # :nodoc:
+    return unless options.is_a?( Hash )
+
+    key = Array( options[:key] ).join( ', ' )
+    raise ArgumentError, 'Attribute :key not found' if key.empty?
```
zdennis: It looks like `:key` represents the [ON CONFLICT](http://www.postgresql.org/docs/9.5/static/sql-insert.html#SQL-ON-CONFLICT)  `conflict_target`. If that's the case can we make this optional?

lib/activerecord-import/import.rb
```diff
@@ -225,13 +225,31 @@ def support_setting_primary_key_of_imported_objects?
...
+    #
+    # The :on_duplicate_key_update option is a hash with :key and
+    # :columns attributes. Below is an example:
+    #
+    #   BlogPost.import columns, values, :on_duplicate_key_update=>{ :key => :id, :columns => [ :date_modified, :content, :author ] }
+    #
+    # ==== :key
+    #
+    # The :key attribute specifies the columns that make up the unique constraint and can be
+    # a single column or an array of column names.
```
zdennis: WDYT about specifically mentioning what key maps to in PostgreSQL?

lib/activerecord-import/import.rb
```diff
@@ -225,13 +225,31 @@ def support_setting_primary_key_of_imported_objects?
...
     # model. Below is an example:
     #
     #   BlogPost.import columns, attributes, :on_duplicate_key_update=>{ :title => :title }
     #
+    # == On Duplicate Key Update (Postgres 9.5+)
+    #
+    # The :on_duplicate_key_update option is a hash with :key and
+    # :columns attributes. Below is an example:
+    #
+    #   BlogPost.import columns, values, :on_duplicate_key_update=>{ :key => :id, :columns => [ :date_modified, :content, :author ] }
```
zdennis: I wonder if we should provide additional example(s) for `:key` usage. The ones that come to mind are:

* Default: no `key` specified. (If this ends up being applicable)
* Passing `key` to use existing index
* Passing `key` to use existing column(s) as index? (I think this is what you have above)

There may be other options such as passing in PG-compatible `index_expression`(s) but I'm not entirely sure if that will already work or if it'd be more work (which I don't think should block this PR getting in).

test/support/postgresql/import_examples.rb
```diff
@@ -127,3 +127,174 @@ def should_support_postgresql_import_functionality
...
+
+        it "should raise an ArgumentError" do
+          assert_raise ArgumentError do
+            perform_import
+          end
+        end
+      end
+    end
+  end
+end
```
zdennis: @jkowens, do you think we can make the default behavior and interface for PostgreSQL compatible with MySQL? If so, I'd love to move all of the baseline examples into a single spec and then leave any examples that are MySQL or PostgreSQL specific (incompatible APIs) into their respective specs. 

Thoughts?

zdennis:
@jkowens, this is great stuff. Great work!  

My main hope before we merge this in and release is it to make the baseline APIs b/w MySQL and PostgreSQL compatible or to share the knowledge of why they cannot be so we can document that reasoning.

I will also work on getting TravisCi to run with PostgreSQL 9.5 so we can have this running by default.

test/postgresql/import_test.rb
```diff
@@ -1,4 +1,7 @@
 require File.expand_path(File.dirname(__FILE__) + '/../test_helper')
 require File.expand_path(File.dirname(__FILE__) + '/../support/postgresql/import_examples')
 
+pg_version = ENV["PG_VERSION"] || "0"
```
zdennis: Reminder to self that I should probably create a test readme file so options like this can be surfaced. 

Out of curiosity what did you use to determine to how to set up and run the test environment? Just figure it out by looking through code or was there any information that was helpful any where in the repos or on the internets?

lib/activerecord-import/adapters/postgresql_adapter.rb
jkowens: Yes, that's correct. The ON CONFLICT clause requires a conflict_target, unless the conflict_action is DO NOTHING. Currently the only time that conflict_action would be DO NOTHING is if :columns is an empty array *and* the table does not have update timestamp columns that automatically get assigned. I can update the code to make :key optional in that scenario

Unfortunately, as far as I can tell, this requirement makes matching the MySQL interface not possible :frowning:

lib/activerecord-import/import.rb
jkowens: Sure thing. Do you think it would be better to rename :key :conflict_target?

test/postgresql/import_test.rb
jkowens: I had to look through the code a bit, but this [wiki entry](https://github.com/zdennis/activerecord-import/wiki/How-to-run-the-tests) really helped.

lib/activerecord-import/adapters/postgresql_adapter.rb
zdennis: Ahh ok. WDYT about defaulting key to :id?

jkowens: Hm I'm wondering if that could lead to more confusion? Maybe someone assumes if it's the primary key they don't need to specify the conflict_target, but it turns out they prefix the id field with table name. Or they expect it to work just like MySQL. I suppose good documentation could remedy that though.

zdennis: What about using AR::Base#primary_key ?

lib/activerecord-import/import.rb
zdennis: :thumbsup: I like that.

lib/activerecord-import/adapters/postgresql_adapter.rb
jkowens: Looks like that should work. I'll make that change so key has a default value (if table has a primary key).

jkowens:
@zdennis I still need to get to documenting options and examples. I will try to do that soon.

lib/activerecord-import/import.rb
jkowens: Could you explain what you mean by passing `key` to use existing index? Postgres has the ability to determine what index maps to a list column names. The Postgres docs say that a constraint name can be explicitly specified, but is not recommended. Should I look into supporting that option?

test/support/postgresql/import_examples.rb
jkowens: @zdennis using your suggestion to default `:conflict_target` to the primary key, it is now possible to provide an array to :on_duplicate_key_update for Postgres.

`:on_duplicate_key_update => [:column1, :column2]`

I've made the array option a baseline example and pulled that out into a separate shared examples file. The hash option is included in both of the adapter specific examples.

lib/activerecord-import/import.rb
jkowens: @zdennis I've added support for explicitly naming the conflicting constraint and I've added an example of that to the comments.

jkowens:
@zdennis hopefully I'm getting closer with these latest updates. Let me know if you have any further suggestions and I will try to get them implemented. I can squash all these commits down into one when the time comes.

lib/activerecord-import/adapters/postgresql_adapter.rb
```diff
@@ -30,6 +31,92 @@ def post_sql_statements( table_name, options ) # :nodoc:
...
+    return unless arg.is_a?( Hash )
+
+    constraint_name = arg[ :constraint_name ]
+    conflict_target = arg[ :conflict_target ]
+    columns         = arg.fetch( :columns, [] )
+
+    sql = " ON CONFLICT "
+    conflict_target = sql_for_conflict_target( conflict_target, constraint_name )
+
+    if columns.respond_to?( :empty? ) && columns.empty?
```
triptec: Is there any test for this block? I've been trying to do something like this:
Category.import categories, :on_duplicate_key_update=>{ :conflict_target => [:name], :columns => [], validate: false }
So if the category exists I don't want to update it, just DO NOTHING

jkowens: Yes, but the code currently looks to see if the table has updated_on or updated_at timestamp columns and if it does those are automatically added to the list of columns, so it's unlikely `columns` will empty. I'm thinking this code was put in place for MySQL since its on duplicate key update clause does not have a "do nothing" option, so columns were required. That brings up a good question, since Postgres does have a "DO NOTHING" option, should we not automatically add the update columns? @zdennis what are your thoughts?

zdennis: @jkowens said:
> Yes, but the code currently looks to see if the table has updated_on or updated_at timestamp columns and if it does those are automatically added to the list of columns, so it's unlikely columns will be empty.

While likely true it's not guaranteed so I agree with @triptec that we should have a test for this.

> I'm thinking this code was put in place for MySQL since its on duplicate key update clause does not have a "do nothing" option, so columns were required

MySQL has INSERT IGNORE which allows I believe achieve the same end-result: ignore inserting rows that fail a required constraint check. This is supported in activerecord-import by passing the `ignore: true` option to `import(...)`. 

In hindsight, this is a terribly named option in activerecord-import.

> That brings up a good question, since Postgres does have a "DO NOTHING" option, should we not automatically add the update columns?

I think we add them if the caller passed in other columns, but only then. See below `on_duplicate_key_ignore` suggestion. 

@triptec said:
> I've been trying to do something like this:
Category.import categories, :on_duplicate_key_update=>{ :conflict_target => [:name], :columns => [], validate: false }

@triptec, so you don't really want to update anything, you just want to ignore duplicate records, yes? If so, what about something like this (which doesn't exist yet):

```
Category.import categories, :on_duplicate_key_ignore=>{ :conflict_target => [:name] }, validate: false
```

This would make it clear that you intend to ignore duplicates and make help us cleanly separate this from the case where we automatically update timestamps when present.

triptec: That would be great.

jkowens: @zdennis @triptec I took a stab at adding `:on_duplicate_key_ignore` for Postgres. As Zach mentioned, MySQL does offer the `ignore: true` option. Passing a hash with a :conflict_target as shown above works, but I thought just passing `true` would more than likely be what most used for this scenario:

```
Category.import categories, on_duplicate_key_ignore: true, validate: false
```

`ON CONFLICT DO NOTHING` is the one time Postgres lets us get away without specifying a conflict target.

test/postgresql/import_test.rb
jkowens: @zdennis I refactored this so that ENV variable PG_VERSION is no longer required. I didn't realize I could grab PostgreSQL version from `ActiveRecord::ConnectionAdapters::PostgreSQLAdapter#postresql_version`.

lib/activerecord-import/adapters/postgresql_adapter.rb
jkowens: I think `:on_duplicate_key_ignore` could potentially be an alias for  MySQL `:ignore`, but I think that probably deserves its own PR.

test/support/shared_examples/on_duplicate_key_update.rb
```diff
@@ -0,0 +1,126 @@
...
+        end
+
+        context "using symbol column names" do
+          let(:update_columns){ [ :title, :author_email_address, :parent_id ] }
+          should_support_on_duplicate_key_update
+          should_update_fields_mentioned
+        end
+      end
+
+      context "given array of model instances with :validation checks turned off" do
```
jkowens: @zdennis is there a need to duplicate tests for importing an array of models to validate :on_duplicate_key_update functionality? If not, I was going to skinny these examples down a bit.

test/support/postgresql/import_examples.rb
zdennis: :thumbsup:

test/support/shared_examples/on_duplicate_key_update.rb
zdennis: I don't know off the top of my head so I'd say go for it. :)

lib/activerecord-import/adapters/postgresql_adapter.rb
```diff
@@ -24,12 +27,113 @@ def next_value_for_sequence(sequence_name)
...
+  end
+
+  # Returns a generated ON CONFLICT DO NOTHING statement given the passed
+  # in +args+.
+  def sql_for_on_duplicate_key_ignore( table_name, *args ) # :nodoc:
+    arg = args.first
+    if arg.is_a?( Hash )
+      conflict_target = sql_for_conflict_target( arg ) if arg.is_a?( Hash )
+    end
+    " ON CONFLICT #{conflict_target}DO NOTHING"
```
zdennis: Does this potentially need a space before `DO NOTHING`?

lib/activerecord-import/import.rb
```diff
@@ -165,15 +165,19 @@ def support_setting_primary_key_of_imported_objects?
...
-    # * +on_duplicate_key_update+ - an Array or Hash, tells import to \
-    #    use MySQL's ON DUPLICATE KEY UPDATE ability. See On Duplicate\
-    #    Key Update below.
+    # * +ignore+ - true|false, tells import to use MySQL's INSERT IGNORE
+    #    to discard records that contain duplicate keys.
+    # * +on_duplicate_key_ignore+ - true|false, tells import to use
+    #    Postgres 9.5+ ON CONFLICT DO NOTHING.
+    # * +on_duplicate_key_update+ - an Array or Hash, tells import to
+    #    use MySQL's ON DUPLICATE KEY UPDATE or Postgres 9.5+ ON CONFLICT
+    #    DO UPDATE ability. See On Duplicate Key Update below.
```
zdennis: TY for updating the docs. I think github ate one of your comments, but I agree with you that `on_duplicate_key_ignore` could become an alias for `ignore`, but it should be done in a separate PR.

lib/activerecord-import/import.rb
```diff
@@ -225,13 +229,73 @@ def support_setting_primary_key_of_imported_objects?
...
+    #
+    #   BlogPost.import columns, values, :on_duplicate_key_update=>{ :conflict_target => :slug, :columns => [ :date_modified, :content, :author ] }
+    #
+    # ========  Using a Hash
+    #
+    # The :columns option can be a hash of column names to model attribute name
+    # mappings. This gives you finer grained control over what fields are updated
+    # with what attributes on your model. Below is an example:
+    #
+    #   BlogPost.import columns, attributes, :on_duplicate_key_update=>{ :conflict_target => :slug, :columns => { :title => :title } }
```
zdennis: Awesome docs above.

lib/activerecord-import/adapters/postgresql_adapter.rb
jkowens: `sql_for_conflict_target` returns a string with a blank space at the end, so I think it will be ok.

zdennis:
Thanks for all your work on this @jkowens! This is really awesome stuff and I know there are a lot of people who are going to be excited for it. The test failure has to do with `em-synchrony` as it doesn't support ActiveRecord 5.0.0 atm so don't worry about that. 

Is there anything else you wanted to do before I merge this in?

jkowens:
Thanks @zdennis :smile: 

If you are good with what we have here, I can squash the commits into one for you to merge.

zdennis:
Let's do it @jkowens

jkowens:
All set :+1:

rwilliams:
Awesome!

zdennis:
Thank you for your work @jkowens ! I will push out a new gem this weekend and PostgreSQL databases everywhere will be filled with the joy of speedier imports from applications using ActiveRecord.

jkowens:
Yes!! :rocket:

fenec:
:fireworks:

zdennis:

Issue #211, opened 2015-12-08
Serialization fails when serializing to JSON


MaxGabriel:
activerecord-import supports saving serialized fields, as described in this spec:

```ruby
  describe "importing serialized fields" do
    it "imports values for serialized fields" do
      assert_difference "Widget.unscoped.count", +1 do
        Widget.import [:w_id, :data], [[1, {:a => :b}]]
      end
      assert_equal({:a => :b}, Widget.find_by_w_id(1).data)
    end
  end
```

However, this fails when `serialize :data, JSON` is used instead of `serialize :data, Hash`. Instead of serializing to JSON, the data is serialized to YAML (the ActiveRecord default). This can be reproduced by changing the `serialize` line in `widget.rb` to `serialize :data, JSON`, as demonstrated [in this branch](https://github.com/zdennis/activerecord-import/compare/master...MaxGabriel:serializeJSON?expand=1).

MaxGabriel:
The tests passed when using `serialize :data, JSON` prior to [this commit](https://github.com/zdennis/activerecord-import/commit/e85ad6ceaa02f5cfc6c01c65461fcdd68c6c37a9), if you also change the serialized fields test to use Strings instead of Symbols (serializing to JSON doesn't preserve the string/symbol distinction).

MaxGabriel:
This problem is potentially limited to Rails 4.1

zdennis:
Thanks for reporting and following up with additional info @MaxGabriel. This appears to work for ActiveRecord 4.0 and higher. I don't see an issue with ActiveRecord 4.1 specifically so if there's a specific version you're concerned with feel free to post that and I will verify.

I've added a test around this with #220 . It will be merged soon.

MaxGabriel:
Hey @zdennis, I forget what the reason was, but I think I concluded this was only limited to our specific version or something. I think this is OK to close.

zdennis:
Thanks @MaxGabriel

elben:
I'm running `activerecord (4.0.13)` (migrating a Rails 3 app) and `activerecord-import (0.10.0)` (I've also tried HEAD d1f6967) and I'm having problems with serialized JSON attributes. The SQL statements look like:

```
class MyModel < ActiveRecord::Base
   serialize :a, JSON
end

[...]

PR #221, opened 2016-01-15
Issues/211 serialize to json

zdennis:
Add tests for the issue reported #211 .

This is build on the work done for #220 . Do not merge until that is merged.


zdennis:
This is no longer necessary based on an update to #211. Closing without adding unnecessary code/tests.

PR #224, opened 2016-01-15
Issues/203 args passed to import should not be modified

### LINKED ISSUE #203: Array passed to `import` should not be modified
When calling `Model.import attrs, records`, the `attrs` array is modified—this is a highly unexpected side effect, and breaks if either `attrs` is frozen, or the program goes on to use `attrs` for something else later.

```
Exception: RuntimeError: can't modify frozen Array
--
 0: activerecord-import-0.10.0/lib/activerecord-import/import.rb:495:in `block in add_special_rails_stamps'
 1: activerecord-import-0.10.0/lib/activerecord-import/import.rb:488:in `each_pair'
 2: activerecord-import-0.10.0/lib/activerecord-import/import.rb:488:in `add_special_rails_stamps'
 3: activerecord-import-0.10.0/lib/activerecord-import/import.rb:300:in `import_helper'
 4: activerecord-import-0.10.0/lib/activerecord-import/import.rb:248:in `import'
```

zdennis:
Resolves #203

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -236,7 +236,7 @@ class ActiveRecord::Base
     # This returns an object which responds to +failed_instances+ and +num_inserts+.
     # * failed_instances - an array of objects that fails validation and were not committed to the database. An empty array if no validation is performed.
     # * num_inserts - the number of insert statements it took to import the data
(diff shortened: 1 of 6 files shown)


zdennis:

PR #226, opened 2016-01-16
Don't directly touch AR::Base, but do it via AS.on_load

zdennis:
Resolves #136


zdennis:
Thanks @amatsuda, merging this in as part of #226.

zdennis:

PR #227, opened 2016-01-16
Don't use AS.on_load for something else than on_load hook

zdennis:
ActiveSupport.on_load is not just a way to add an additional block of code into a Ruby method.
It's a mechanism for adding load hooks which is needed to implement lazy loading in ActiveSupport.

And, in this particular after_established case, we don't actually need any special mechanism. Let's just simply call the logic inside the AMC.

This is to get changes in #177 implemented by @amatsuda  in.


zdennis:

PR #230, opened 2016-02-05
Recursively import all has_one/has_many associations

sferik:
This should work for all `has_one`/`has_many` associations, not just ones explicitly marked `autosave`.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -179,9 +179,9 @@ class ActiveRecord::Base
     #   existing model instances in memory with updates from the import.
     # * +timestamps+ - true|false, tells import to not add timestamps
     #   (if false) even if record timestamps is disabled in ActiveRecord::Base
-    # * +recursive - true|false, tells import to import all autosave association
-    #   if the adapter supports setting the primary keys of the newly imported
-    #   objects.
+    # * +recursive - true|false, tells import to import all has_many/has_one
+    #   associations if the adapter supports setting the primary keys of the
+    #   newly imported objects.
     #
     # == Examples
     #  class BlogPost < ActiveRecord::Base ; end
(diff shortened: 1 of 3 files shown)


lib/activerecord-import/import.rb
```diff
@@ -462,7 +461,7 @@ def import_associations(models, options)
     def find_associated_objects_for_import(associated_objects_by_class, model)
       associated_objects_by_class[model.class.name]||={}
 
-      model.class.reflect_on_all_autosave_associations.each do |association_reflection|
+      model.class.reflect_on_all_associations(:has_many).each do |association_reflection|
```
jkowens: I think this might need to be revised to include :has_one associations.

jkowens:
:+1:

lib/activerecord-import/import.rb
sferik: Yes. That can be added once #234 is merged.

zdennis: @sferik , #234 just got merged in. :)

sferik:
@zdennis I just rebased this branch from `master` and added back support for `has_one` associations.

zdennis:
Awesome! Thanks @sferik!

sferik:
As you’re well aware, the tests are failing due to an [issue with `em-synchrony`](https://github.com/igrigorik/em-synchrony/issues/201), which you reported one month ago and has received no response. That project shows no sign of being updated to work with Active Record 5. According to [this open issue](https://github.com/igrigorik/em-synchrony/issues/198), it still doesn’t work with Active Record 4.2.1, which was released on March 19, 2015. In fact, the last commit to `em-synchrony` was on March 16, 2015, almost a year ago.

At this point, I think dropping support for `em-synchrony` is the right thing to do for this project. If someone wants to use `activerecord-import` with `em-synchrony`, they can take ownership over getting the test build passing with it. In the mean time, I will open a pull request that removes `em-synchrony` from the test suite and turns the build green. You can merge it if/when you agree with this assessment.

zdennis:
Closing this in favor of #243 which builds on this PR. Thanks for your work on this @sferik.

Also `em-synchrony` and `mysql` have been removed in #239.

zdennis:

PR #234, opened 2016-02-15
Fix issue importing has_one associations

### LINKED ISSUE #190: Private method called when importing has_one relationships
Consider the following model, with a Question having one Rule:
```ruby
class Question < ActiveRecord::Base
	has_one :rule, autosave: true
end
```
```ruby
class Rule < ActiveRecord::Base
	belongs_to :question
end
```
Calling Question#import (with `recursive: true`) yields the following:
```
<#<NoMethodError: private method `select' called for #<ActiveRecord::Associations::HasOneAssociation:0x007fc26784a658>>>.
```

jkowens:
Fixes #190.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -532,6 +532,9 @@ class ActiveRecord::Base
         association = model.association(association_reflection.name)
         association.loaded!
 
+        # Wrap target in an array if not already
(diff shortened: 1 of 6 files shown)


zdennis:
:100: thanks for finding and fixing this @jkowens

zdennis:

PR #235, opened 2016-02-17
Fix rails (> 4.1) enum column with a default integer value

aquajach:
First, _ENV['AR_VERSION'].to_i > 4.1_ is only true when _AR_VERSION >= 5_ because _to_i_ rounds down strings, e.g. _"4.2".to_i => 4_. Therefore, in 4.2 the enum in book.rb is not currently defined.

And the main stuff is, for many adapters the default value of an integer column is still in string format (like in postgresql by using pg_get_expr, check at: http://www.postgresql.org/docs/8.0/static/functions-info.html#AEN13319) before type casting. So during the importing, an argument error happens like _ArgumentError: '0' is not a valid status_, given 0 is the default value for status column. This issue seems not only to happen in postgresql.

There's no new test cases for this PR, because some existing tests get failed when default value is added and enum is enabled in book.rb. Certainly, the implement fixes all of them.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -275,7 +275,8 @@ class ActiveRecord::Base
           # this next line breaks sqlite.so with a segmentation fault
           # if model.new_record? || options[:on_duplicate_key_update]
             column_names.map do |name|
-              if ActiveRecord::Base.connection.instance_of?(ActiveRecord::ConnectionAdapters::PostgreSQLAdapter)
+              env =  ENV['RAILS_ENV'] || Rails.env
+              if ActiveRecord::Base.configurations[env]["adapter"] == 'postgresql'
                 model.read_attribute(name.to_s)
               else
                 model.read_attribute_before_type_cast(name.to_s)
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -276,7 +276,7 @@ class ActiveRecord::Base
           # if model.new_record? || options[:on_duplicate_key_update]
             column_names.map do |name|
               env =  ENV['RAILS_ENV'] || Rails.env
-              if ActiveRecord::Base.configurations[env]["adapter"] == 'postgresql'
+              if %w{postgresq em_mysql2}.include? ActiveRecord::Base.configurations[env]["adapter"]
                 model.read_attribute(name.to_s)
               else
                 model.read_attribute_before_type_cast(name.to_s)
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -275,8 +275,7 @@ class ActiveRecord::Base
           # this next line breaks sqlite.so with a segmentation fault
           # if model.new_record? || options[:on_duplicate_key_update]
             column_names.map do |name|
-              env =  ENV['RAILS_ENV'] || Rails.env
-              if %w{postgresq em_mysql2}.include? ActiveRecord::Base.configurations[env]["adapter"]
+              if model.class.column_defaults[name.to_s].is_a? Integer
                 model.read_attribute(name.to_s)
               else
                 model.read_attribute_before_type_cast(name.to_s)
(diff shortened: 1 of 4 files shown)


aquajach:
It seems the only thing making the build failed is about em-synchrony in ar >= 5.0 now

lib/activerecord-import/import.rb
```diff
@@ -275,7 +275,11 @@ def import_helper( *args )
           # this next line breaks sqlite.so with a segmentation fault
           # if model.new_record? || options[:on_duplicate_key_update]
             column_names.map do |name|
-              model.read_attribute_before_type_cast(name.to_s)
+              if model.class.column_defaults[name.to_s].is_a? Integer
+                model.read_attribute(name.to_s)
+              else
+                model.read_attribute_before_type_cast(name.to_s)
```
jkowens: Can someone why `read_attribute_before_type_cast` is being called instead of just `read_attribute` anyway?

test/import_test.rb
```diff
@@ -404,7 +404,7 @@
       end
     end
 
-    if ENV['AR_VERSION'].to_i > 4.1
+    if ENV['AR_VERSION'].to_f > 4.1
```
jkowens: Nice catch!

lib/activerecord-import/import.rb
zdennis: The original commit was back in early 2007: https://github.com/zdennis/ar-extensions/commit/b993d23533e8342d856377fd77006127679ba569

The commit message kind of sucks so I apologize for that. 

IIRC, it's because we are collecting DB-safe representations of the values and ActiveRecord may give you back an object that's nice to work with in Ruby but that may not be db-safe. I think Date, DateTime, and boolean fields were an example of this.

But, if we can remove it and have the test suite continue to pass with just `read_attribute` I'm all for it.

jkowens: Ah ok, that makes sense. It looks like the method `values_sql_for_columns_and_attributes` ultimately handles casting attributes to db safe values now? I think it just using  `read_attribute` might be ok.

@aquajach would you be able to try that out?

aquajach: `read_attribute ` is slightly slower than `read_attribute_before_type_cast ` for type casting. 

But for rails enum, type casting happens **after** enum value checking. For example, `Person.new(age: "10")` is fine even if age column is an integer, but `Person.new(status: "1")` will raise `ArgumentError: '1' is not a valid status` given status is an enum integer attribute.

One thing I just discovered, `model.class.column_defaults` is very slow. So I am going to make another commit to cache it or something.

aquajach: @jkowens ok will try before the next commit

aquajach: @jkowens @zdennis  Meanwhile, I did the benchmark for `read_attribute_before_type_cast ` and `read_attribute ` which tells replacing `read_attribute_before_type_cast`  with `read_attribute` seems not be very ideal:

https://gist.github.com/aquajach/f875e13773ace44d6f85

jkowens:
It looks like issue #140 could be a similar issue (validations failing because of using `read_attribute_before_type_cast`). I think the fix suggested by @trandoanhung1991 might be worth investigating. If it works out there might even be a small performance boost.

lib/activerecord-import/import.rb
jkowens: Well that certainly explains why `read_attribute_before_type_cast` was used :wink: 

/cc @zdennis

test/import_test.rb
zdennis: :thumbsup:

zdennis:
I'm going to merge this in for now. The issue @trandoanhung1991 identified can be handled separately and should get a test for that specific case. We can open another PR if the fix for that simplifies or removes the `read_attribute_before_type_cast` call.

aquajach:
@zdennis thanks, I have checked that #237 fixing it in a better approach without using `read_attribute`

zdennis:

aquajach:
👍

PR #238, opened 2016-02-22
Remove tests for em_mysql2 adapter

sferik:
It appears that `em-synchrony` is no longer being updated and no longer works with recent versions of Active Record, causing the build to fail. This commit can be reverted when these `em-synchrony` issues are resolved.

Please see [this comment](https://github.com/zdennis/activerecord-import/pull/230#issuecomment-187023507) for additional rationale.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,7 +13,7 @@ namespace :display do
 end
 task :default => ["display:notice"]
 
-ADAPTERS = %w(mysql mysql2 em_mysql2 jdbcmysql jdbcpostgresql postgresql sqlite3 seamless_database_pool mysqlspatial mysql2spatial spatialite postgis)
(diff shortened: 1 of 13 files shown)


sferik:
There appear to be some build legitimate errors that may have been going undetected up to this point due to the `em-synchrony` and `mysql` related build failures, which were masking these errors.

zdennis:
Thanks for doing this @sferik, I've created PR #239 based on your work here since I needed to remove a few more things to rid activerecord-import of `em-synchrony` and `mysql`.

For future readers to this issue: **this removes support the em-synchrony and mysql adapters**. For `em-synchrony` there is no replacement as it appears abandoned. For `mysql` please use the `mysql2` gem

zdennis:

PR #242, opened 2016-02-26
Updating travisci build to be less verbose on successful runs.

zdennis:
This makes the output on TravisCI look like this in the success case:

screenshot 2016-02-26 13 27 41

In the failure case it will show the full output for which run failed.


jkowens:
Nice!

PR #245, opened 2016-03-02
Add :batch_size option to import

### LINKED ISSUE #207: Add support for import batch size
Using `import`, I brought down our QA server when it tried to insert 500,000 records in a single INSERT statement.

Of course I can fix this myself externally, but a batch size option with a sensible default would be nice!

By the way, thanks for the great work.. activerecord-import is giving us about a 10x speedup on our import tasks.

jkowens:
This resolves #207. Batch size defaults to the total number of records to import. I considered using 1000 records per insert as a default as that seems to be fairly standard, but I decided it might be better to have it as an explicit option for now. I'm open to other suggestions :smile: 

See [ActiveRecord::Batches](http://api.rubyonrails.org/classes/ActiveRecord/Batches.html).

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -178,9 +178,11 @@ class ActiveRecord::Base
     #   existing model instances in memory with updates from the import.
     # * +timestamps+ - true|false, tells import to not add timestamps
     #   (if false) even if record timestamps is disabled in ActiveRecord::Base
-    # * +recursive - true|false, tells import to import all has_many/has_one
+    # * +recursive+ - true|false, tells import to import all has_many/has_one
     #   associations if the adapter supports setting the primary keys of the
     #   newly imported objects.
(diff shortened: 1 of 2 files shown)


sferik:
@jkowens @zdennis Just merged a pull request that cleaned-up the styling across a large number of files. This has caused a conflict between `master` and your patch. Would you mind pulling from this repo and rebasing from `master`?

jkowens:
@sferik @zdennis just an FYI that I have rebased master.

zdennis:
@jkowens, thanks for adding this. 

I usually run it in batches of 25_000 and I imagine that everyone else as a different setting as it heavily depends on how much memory is available and that may be system dependent. Unlike `find_in_batches` massive imports likely don't run inside of a web request/response cycle so I think avoiding the default of 1_000 makes sense. And for those that do run inside of a web request/response cycle those are usually very limited and size and wouldn't be affected anyways.

PR #246, opened 2016-03-02
Add import! method

### LINKED ISSUE #162: import!
Similar to AR `create!`, it would be nice to have a variant of import, `import!`, which raised a `RecordInvalid` exception if validation errors occurred.

jkowens:
When using `import!` an ActiveRecord::RecordInvalid error is raised if any of the validations fail. Resolves #162.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -312,6 +312,16 @@ class ActiveRecord::Base
       end
     end
 
+    # Imports a collection of values if all values are valid. Import fails at the
+    # first encountered validation error and raises ActiveRecord::RecordInvalid
+    # with the failed instance.
+    def import!(*args)
+      options = args.last.is_a?( Hash ) ? args.pop : {}
+      options.merge!( { validate: true, raise_error: true } )
+
+      import(*args, options)
+    end
+
     def import_helper( *args )
       options = { validate: true, timestamps: true, primary_key: primary_key }
       options.merge!( args.pop ) if args.last.is_a? Hash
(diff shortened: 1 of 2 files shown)


test/import_test.rb
```diff
@@ -461,4 +461,28 @@
...
+    end
+
+    context "with valid data" do
+      it "should import data" do
+        assert_difference "Topic.count", +2 do
+          Topic.import! columns, valid_values
+        end
+      end
+    end
+  end
```
zdennis: :thumbsup:

lib/activerecord-import/import.rb
```diff
@@ -313,6 +313,14 @@ def import(*args)
       end
     end
 
+    # Import if all values are valid, otherwise raise an error
```
zdennis: WDYT about mentioning the specific error being raised here and the fact that this will fail-fast? I know the not bang version of this will collect the failed instances and return them all.

jkowens: This is a good suggestion :smile: I will work on improving this documentation.

sferik:
@jkowens @zdennis just merged a pull request that cleaned-up the styling across a large number of files. This has caused a conflict between `master` and this pull request. Would you mind pulling from this repo and rebasing from `master`?

jkowens:
@sferik thanks for the heads up. I rebased this with master so it can be merged.

lib/activerecord-import/import.rb
```diff
@@ -312,6 +312,16 @@ def import(*args)
...
 
+    # Imports a collection of values if all values are valid. Import fails at the
+    # first encountered validation error and raises ActiveRecord::RecordInvalid
+    # with the failed instance.
+    def import!(*args)
+      options = args.last.is_a?( Hash ) ? args.pop : {}
+      options.merge!( { validate: true, raise_error: true } )
+
+      import(*args, options)
+    end
```
zdennis: :thumbsup:

PR #252, opened 2016-03-17
Add support for polymorphic associations

sferik:


sferik:
Looks like this one requires a rebase as well. I will do that now.

sferik:
@zdennis Done.

sferik:
@zdennis Done.

lib/activerecord-import/import.rb
```diff
@@ -421,7 +421,7 @@ def import_with_validations( column_names, array_of_attributes, options = {} )
       # validation we'll use the index to remove it from the array_of_attributes
       arr.each_with_index do |hsh, i|
         instance = new do |model|
-          hsh.each_pair { |k, v| model.send("#{k}=", v) }
+          hsh.each_pair { |k, v| model.public_send("#{k}=", v) }
```
zdennis: :100: on the usage of `public_send` to respecting its visibility .

lib/activerecord-import/import.rb
```diff
@@ -542,7 +542,11 @@ def find_associated_objects_for_import(associated_objects_by_class, model)
 
         changed_objects = association.select { |a| a.new_record? || a.changed? }
         changed_objects.each do |child|
-          child.send("#{association_reflection.foreign_key}=", model.id)
+          child.public_send("#{association_reflection.foreign_key}=", model.id)
+          # For polymorphic associations
+          association_reflection.type.try do |type|
+            child.public_send("#{type}=", model.class.name)
+          end
```
zdennis: :thumbsup: I've used `try` a million times and never used it's block syntax. Good stuff.

test/models/discount.rb
```diff
@@ -0,0 +1,3 @@
+class Discount < ActiveRecord::Base
+  belongs_to :discountable, polymorphic: true
```
zdennis: Thank you for adding a test and updating the test fixtures! This makes it so much easier to review the changes.

zdennis:
@sferik, I'm happy to merge this in. I can do so before or after PR #256. Let me know if you want this or that to come first.

zdennis:
Actually, I'm just going to merge this in and see what happens with PR #256. Living courageously.

lib/activerecord-import/import.rb
sferik: Yeah, it’d be bad if we invoked a private method here without raising.

PR #254, opened 2016-03-18
Drop support for Rails 3.1

sferik:
Rails 3.1 is no longer being maintained, [even for severe security issues](http://guides.rubyonrails.org/maintenance_policy.html). The last 3.1.x release, [was 3 years ago](https://rubygems.org/gems/activerecord/versions).

This will also make the Travis CI build run significantly faster, since we can run 5 builds in parallel and this is the 6th build, so it has to wait for all the other Active Record versions to finish before it can start.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ env:
     # https://github.com/discourse/discourse/blob/master/.travis.yml
     - RUBY_GC_MALLOC_LIMIT=50000000
   matrix:
-    - AR_VERSION=3.1
     - AR_VERSION=3.2
     - AR_VERSION=4.0
     - AR_VERSION=4.1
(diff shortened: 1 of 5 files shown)


zdennis:
I like the idea of removing support for ActiveRecord 3.1. Unless there are obstacles in the way for supporting it for the next release I may want to remove this after the next `activerecord-import` release, so in 2 releases it would be gone. 

But, if there are any challenges I'm fine to nix 3.1 sooner.

Thoughts?

sferik:
@zdennis That seems totally reasonable. I’ll just leave this pull request open until you’re ready to merge it.

zdennis:
Bye bye Rails 3.1 thanks to @sferik!

zdennis:

PR #255, opened 2016-03-18
Fix line endings of benchmark files

sferik:
For some reason, many of the files in the `benchmark` directory have non-Unix line endings. This patch makes these files consistent with the rest of the project.


sferik:
Superseded by https://github.com/zdennis/activerecord-import/pull/256.

PR #256, opened 2016-03-18
Add RuboCop to enforce style rules

sferik:
This is a follow-up to https://github.com/zdennis/activerecord-import/pull/250. It does a bit more cleanup and adds RubyCop to the test suite, so the build will fail there are any style guide regressions (as specified in `.rubocop.yml`).

It also adds the following two `rake` tasks:

```
rake rubocop                      # Run RuboCop
rake rubocop:auto_correct         # Auto-correct RuboCop offenses
```

The first task is the one that runs on Travis CI and will report any violations of the style guide. The second is intended to be run locally and will attempt to fix any style guide violations that can be automatically corrected (remove trailing whitespace, convert tabs into spaces, etc.)

If you’d like, I can also add a few sentences in the `README` or create a `CONTRIBUTING` doc that informs potential developers about the style guide and how to adhere to it.

This will ensure a consistent style is used throughout this project, which should make the code easier to work with going forward.

PR #257, opened 2016-03-18
Fix issue doing recursive import with :on_duplicate_key_update

### LINKED ISSUE #249: Error on using :recursive and :on_duplicate_key_update parameters together
```
Product.import(products, recursive: true, :on_duplicate_key_update=>{ :constraint_name => 'products_pkey', :columns => [ :internal_number, :site_id] })

=> PG::UndefinedObject: ERROR:  constraint "products_pkey" for table "attributes" does not exist
: INSERT INTO "attributes" ("id","price","uom_price","promo_price","promo_uom_price","shipping_cost","level","stock","day_or_week","product_id") VALUES ...
```

The code is searching the constraint **products_pkey** in wrong table **attributes** when the constraint exists in **products** table. I think the straight-forward fix is to apply :on_duplicate_key_update parameter values only for parent model (for Products). 
Anyway it would be perfect it you can fix it for `child models` as well.

jkowens:
Only use :on_duplicate_key_update option when importing top level models. Fixes #249.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -394,7 +394,7 @@ class ActiveRecord::Base
         set_ids_and_mark_clean(models, return_obj)
 
         # if there are auto-save associations on the models we imported that are new, import them as well
-        import_associations(models, options) if options[:recursive]
+        import_associations(models, options.dup) if options[:recursive]
       end
 
       return_obj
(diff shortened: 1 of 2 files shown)


zdennis:
Thank you @jkowens for the PR and @sferik for the review. 

@sferik, I'm going to jump ahead of PR #256 and merge this in.

PR #237, opened 2016-02-21
Assign attributes directly during import validation

trandoanhung1991:
To prevent parsing UTC time into local datetime.

PR #259, opened 2016-03-22
Refactor attribute assignment for models

jkowens:
This simplifies logic for fetching and assigning attribute values for models. Includes code from #237.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -345,18 +345,16 @@ class ActiveRecord::Base
         end
 
         array_of_attributes = models.map do |model|
-            # this next line breaks sqlite.so with a segmentation fault
-            # if model.new_record? || options[:on_duplicate_key_update]
-            column_names.map do |name|
-              name = name.to_s
-              if respond_to?(:defined_enums) && defined_enums.key?(name) # ActiveRecord 5
-                model.read_attribute(name)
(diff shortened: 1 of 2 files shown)


lib/activerecord-import/import.rb
```diff
@@ -335,18 +335,16 @@ def import_helper( *args )
         end
 
         array_of_attributes = models.map do |model|
-            # this next line breaks sqlite.so with a segmentation fault
-            # if model.new_record? || options[:on_duplicate_key_update]
-            column_names.map do |name|
-              name = name.to_s
-              if respond_to?(:defined_enums) && defined_enums.key?(name) # ActiveRecord 5
-                model.read_attribute(name)
-              elsif model.class.column_defaults[name].is_a?(Integer)
```
zdennis: I think I let this change slip through without a test in 511533450b69c0755fae0f2faf493a7d7356c3d1 . If you're able to add a test for this @jkowens that would be great. If not let me know and I'll put it on my docket. I want to make sure we don't back out an actual fix for a bug fix.

jkowens: @zdennis sorry I didn't realize there wasn't a test added for that specific issue. That commit fixed the issue when an enum attribute had a default value set on the database column. I've added a test case for that.

zdennis: :100: @jkowens ! Gracias!

PR #264, opened 2016-04-01
Fix benchmark html report

jkowens:
The shovel operator has higher precedence than the equality operator.


zdennis:

PR #266, opened 2016-04-04
Improve performance of importing models with validations

jkowens:
Currently, when validating models on import, the models are converted to arrays of values and then are reinitialized as models to call `valid?` It seems unnecessary to do this extra reinitialization step. Inspired by #118.

Running the MySQL benchmarks on my laptop shows a 2x performance increase (over previous implementation) for importing models with validations.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -350,12 +350,7 @@ class ActiveRecord::Base
           # this next line breaks sqlite.so with a segmentation fault
           # if model.new_record? || options[:on_duplicate_key_update]
           column_names.map do |name|
-            name = name.to_s
-            if respond_to?(:defined_enums) && defined_enums.key?(name) # ActiveRecord 5
-              model.read_attribute(name)
-            else
-              model.read_attribute_before_type_cast(name)
-            end
+            model.read_attribute_before_type_cast(name.to_s)
           end
           # end
         end
@@ -387,7 +382,19 @@ class ActiveRecord::Base
       end
 
       return_obj = if is_validating
-        import_with_validations( column_names, array_of_attributes, options )
+        if models
+          import_with_validations( column_names, array_of_attributes, options ) do |failed|
+            models.each_with_index do |model, i|
+              model = model.dup if options[:recursive]
+              next if model.valid?(options[:validate_with_context])
+              model.send(:raise_record_invalid) if options[:raise_error]
+              array_of_attributes[i] = nil
+              failed << model
+            end
+          end
+        else
+          import_with_validations( column_names, array_of_attributes, options )
+        end
       else
         (num_inserts, ids) = import_without_validations_or_callbacks( column_names, array_of_attributes, options )
         ActiveRecord::Import::Result.new([], num_inserts, ids)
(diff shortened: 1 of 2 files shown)


lib/activerecord-import/import.rb
```diff
@@ -387,7 +387,19 @@ def import_helper( *args )
       end
 
       return_obj = if is_validating
-        import_with_validations( column_names, array_of_attributes, options )
+        if models
+          import_with_validations( column_names, array_of_attributes, options ) do |failed|
+            models.each_with_index do |model, i|
+              model = model.dup if options[:recursive]
```
jkowens: Cloning the model was necessary when the `:recursive` option was set so that the parent model would be imported even if child associations failed validation. This is apparently the desired behavior since it is described [in this test](https://github.com/zdennis/activerecord-import/blob/master/test/support/postgresql/import_examples.rb#L108).

jkowens: Duping the model was necessary when the `:recursive` option was set so that the parent model would be imported even if child associations failed validation. This is apparently the desired behavior since it is described [in this test](https://github.com/zdennis/activerecord-import/blob/master/test/support/postgresql/import_examples.rb#L108).

jkowens:
Apparently there is a known bug using `dup` with Ruby 2+ and ActiveRecord 3.1. [Rails 3.2.13](http://weblog.rubyonrails.org/2013/2/27/Rails-3-2-13-rc1-has-been-released/) was the first version to officially support Ruby 2.0, hence the failing test.

This will be a moot point once #254 is merged :smile:

lib/activerecord-import/import.rb
```diff
@@ -350,12 +350,7 @@ def import_helper( *args )
           # this next line breaks sqlite.so with a segmentation fault
           # if model.new_record? || options[:on_duplicate_key_update]
           column_names.map do |name|
-            name = name.to_s
-            if respond_to?(:defined_enums) && defined_enums.key?(name) # ActiveRecord 5
```
jkowens: This logic doesn't seem to be necessary if models aren't converted to arrays and then back to models.

lib/activerecord-import/import.rb
```diff
@@ -424,21 +431,26 @@ def import_from_table( options ) # :nodoc:
...
+      if block_given?
+        yield failed_instances
+      else
+        # create instances for each of our column/value sets
+        arr = validations_array_for_column_names_and_attributes( column_names, array_of_attributes )
+
+        # keep track of the instance and the position it is currently at. if this fails
+        # validation we'll use the index to remove it from the array_of_attributes
+        arr.each_with_index do |hsh, i|
+          instance = new do |model|
```
jkowens: @zdennis instead of instantiating a new instance for every record, could we instantiate one model and then just update all the values on it for each record before validating? I that should give a nice performance boost.

lib/activerecord-import/import.rb
```diff
@@ -424,21 +431,24 @@ def import_from_table( options ) # :nodoc:
...
 
-      # keep track of the instance and the position it is currently at. if this fails
-      # validation we'll use the index to remove it from the array_of_attributes
-      arr.each_with_index do |hsh, i|
-        instance = new do |model|
+        # keep track of the instance and the position it is currently at. if this fails
+        # validation we'll use the index to remove it from the array_of_attributes
+        model = new
+        arr.each_with_index do |hsh, i|
           hsh.each_pair { |k, v| model[k] = v }
```
jkowens: @zdennis instead of instantiating a new instance for every record, here I changed it to instantiate one instance and just update the attributes on that model.

zdennis: Very nice!

zdennis:
Thanks @jkowens !

zdennis:
This will go out in the 0.14.0 release.

PR #268, opened 2016-04-05
Prevent :on_duplicate_key_ignore being used when :recursive enabled

### LINKED ISSUE #267: Postgres :on_duplicate_key_ignore not compatible with :recursive option
Postgres `ON CONFLICT DO NOTHING` does not return primary keys for duplicate records, so the `:recursive` option does not work in combination with it the `:on_duplicate_key_ignore` option.

Any suggestion on how to handle these mutually exclusive options? Should an error be raised if both are enabled or should one of these options take precedence?

jkowens:
Recursive import depends on the primary keys of the parent model being returned, but :on_duplicate_key_ignore doesn't return ids for duplicate records. It is possible then, that a model will be assigned the wrong id and its children would be associated with the wrong parent.

Fixes #267.


zdennis:
:thumbsup: thank you @jkowens!

zdennis:

PR #269, opened 2016-04-05
Only attempt recursive import when passed array of models

jkowens:
Prevents a `NoMethodError: undefined method 'each' for nil:NilClass` if :recursive option was erroneously set when importing with an array of columns and an array of values.

Instead the :recursive option is ignored and the import is completed.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -418,7 +418,7 @@ class ActiveRecord::Base
       return_obj.num_inserts = 0 if return_obj.num_inserts.nil?
 
       # if we have ids, then set the id on the models and mark the models as clean.
-      if support_setting_primary_key_of_imported_objects?
+      if models && support_setting_primary_key_of_imported_objects?
         set_ids_and_mark_clean(models, return_obj)
 
         # if there are auto-save associations on the models we imported that are new, import them as well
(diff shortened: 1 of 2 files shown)


zdennis:
@jkowens, WDYT about raising an error if `recursive` is erroneously provided?

jkowens:
@zdennis I think the "parent" values would have already been inserted into the database at this point...so raising an error might cause more confusion, since the import was actually successful.

zdennis:
> I think the whole process would need to be wrapped in a transaction if we were to raise an error here in order to keep it atomic. What are your thoughts?

I think I'd like to continue to push transactions down to the caller at this point. I'm being a little cautious as to taking on ownership of large transactions.

lib/activerecord-import/import.rb
```diff
@@ -529,6 +529,7 @@ def import_associations(models, options)
       # notes:
       #    does not handle associations that reference themselves
       #    should probably take a hash to associations to follow.
+      return if models.nil?
```
zdennis: Do you know of a situation when `models` will be `nil`? I tried to write a test for this locally, but I couldn't see to find the right invocation to make it so.

jkowens: Really the only time is if import is passed an array of columns and values, which is an invalid/unlikely scenario.

```
columns = [ :title, :author ]
values = [ ['Book1', 'FooManChu'], ['Book2', 'Bob Jones'] ]
Book.import columns, values, validate: false, recursive: true
```

jkowens:
@zdennis I've added a test to this commit.

zdennis:
Thanks @jkowens

PR #271, opened 2016-04-17
Remove deprecated use of `alias_method_chain` in Rails 5

### LINKED ISSUE #270: Use of alias_method_chain has been depreciated in Rails 5
This is my first time using Github / commenting on a project, so forgive me if I've made a mistake, but while upgrading my Rails 4.2 app to Rails 5, I've encountered a depreciation warning related to activerecord-import's use of `alias_method_chain` found in lib > activerecord-import.rb.

As detailed here in [Riding on Rails](http://weblog.rubyonrails.org/2015/3/27/this-week-in-rails-goodbye-alias_method_chain-postgresql-typecasting-and-more/), and here in this Rails pull [request 19434](https://github.com/rails/rails/pull/19434), `alias_method_chain` has been depreciated in favor of Module.prepend. I've [looked into](http://www.justinweiss.com/articles/rails-5-module-number-prepend-and-the-end-of-alias-method-chain/) offering up a fix using Module.prepend, but I'm afraid I'm just not knowledgeable enough in Ruby.

codeodor:
This would close #270

### DIFF
--- a/lib/activerecord-import.rb
+++ b/lib/activerecord-import.rb
@@ -12,7 +12,8 @@ ActiveSupport.on_load(:active_record) do
         ActiveRecord::Import.load_from_connection_pool connection_pool
         conn
       end
-      alias_method_chain :establish_connection, :activerecord_import
+      alias establish_connection_without_activerecord_import establish_connection
+      alias establish_connection establish_connection_with_activerecord_import
     end
   end
 end


codeodor:
Hold off on merging this. All the tests pass for me, but when using it in my app's tests, we get infinite recursion on establishing a connection.

codeodor:
Well, running it in dev mode it works fine, and even running the single failing test works fine. So this is might take a while to run down.

codeodor:
Ok, after some research, I could not figure out what was going wrong, so I switched gears and just used `alias_method`.

It should be safe now. 😁

zdennis:

PR #273, opened 2016-04-25
Remove redundant conditions

pavlik4k:


lib/activerecord-import/import.rb
```diff
@@ -361,7 +361,7 @@ def import_helper( *args )
         end
         # supports empty array
       elsif args.last.is_a?( Array ) && args.last.empty?
-        return ActiveRecord::Import::Result.new([], 0, []) if args.last.empty?
+        return ActiveRecord::Import::Result.new([], 0, [])
```
zdennis: Thanks @pavlik4k !

PR #274, opened 2016-05-08
Mysql2Makara support

bf4:
Perhaps we should make it easier to add behavior like this outside of a fork via template method?

### DIFF
--- a/lib/activerecord-import/base.rb
+++ b/lib/activerecord-import/base.rb
@@ -7,6 +7,7 @@ module ActiveRecord::Import
 
   def self.base_adapter(adapter)
     case adapter
+    when 'mysql2_makara' then 'mysql2'
     when 'mysql2spatial' then 'mysql2'
     when 'spatialite' then 'sqlite3'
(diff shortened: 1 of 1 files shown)


jkowens:
Could `postgresql_makara` support be added in with this PR too?

jkowens:
I noticed there is also PR #204 that uses another method of creating a proxy adapter. @zdennis which way is preferred?

zdennis:
I prefer what @bf4 has here. If we can re-use an existing adapter as-is might as well. It's easy enough to break out if the functionality ever needs to be customized.

We should make sure that we add this to https://github.com/zdennis/activerecord-import/blob/master/Rakefile#L16 and https://github.com/zdennis/activerecord-import/blob/master/.travis.yml#L40 so the test suite runs against it.

zdennis:
Thanks @bf4 and @jkowens. I made some updates locally and merged into master in c7424fd

bf4:
Yay!

PR #276, opened 2016-05-16
Added no_returning option for Postgres to insert without RETURNING clause

makaroni4:
Motivation for this PR: enable import without `RETURNING` clause (right now for every INSERT statement `RETURNING` clause is applied). This will allow to use Postgres rules for insertion, for example the rule to simply ignore all duplications inside INSERT statement.

Right now if we have smth like this in DB (Postgres younger than 9.5)

~~~
CREATE OR REPLACE RULE events_duplicate_ignore AS ON insert TO events
  WHERE EXISTS(
    SELECT 1
    FROM events
    WHERE (mask)=(NEW.mask)
  )
  DO INSTEAD NOTHING;
~~~

and try to insert duplicate we'll have an error `ActiveRecord::StatementInvalid: PG::FeatureNotSupported: ERROR: cannot perform INSERT RETURNING on relation "events" HINT: You need an unconditional ON INSERT DO INSTEAD rule with a RETURNING clause.`

### DIFF
--- a/lib/activerecord-import/adapters/postgresql_adapter.rb
+++ b/lib/activerecord-import/adapters/postgresql_adapter.rb
@@ -26,7 +26,7 @@ module ActiveRecord::Import::PostgreSQLAdapter
   end
 
   def post_sql_statements( table_name, options ) # :nodoc:
-    if options[:primary_key].blank?
+    if options[:no_returning] || options[:primary_key].blank?
       super(table_name, options)
     else
       super(table_name, options) << "RETURNING #{options[:primary_key]}"
--- a/test/support/postgresql/import_examples.rb
+++ b/test/support/postgresql/import_examples.rb
@@ -147,6 +147,20 @@ def should_support_postgresql_import_functionality
         end
       end
     end
+
+    describe "no_returning" do
+      let(:books) { [Book.new(author_name: "foo", title: "bar")] }
+
+      it "creates records" do
+        assert_difference "Book.count", +1 do
+          Book.import books, no_returning: true
+        end
+      end
+
+      it "returns no ids" do
+        assert_equal [], Book.import(books, no_returning: true).ids
+      end
+    end
   end
 end
 


jkowens:
@makaroni4 I believe the `RETURNING` clause was added to make `:recursive` import possible. Instead of a new option `:no_returning`, I'm wondering if you could just check and see if `:recursive` is enabled? If not skip appending that clause.

makaroni4:
Hi Jordan!

Thanks for the quick response! I think there is even more general case –
import should perform like 1 insert. So RETURNING enables you to work with
new entries right after you inserted them. But for import I don't think
it's usually the case – the ultimate goal is to just have this records in
db (like any analytics data). What do you think?

Btw I used my branch in our production setup for importing data – works
super smooth, the rule in postgres just ignores duplicated data on insert.

Best,
Anatoli

jkowens:
After more thought I decided your explicit option is the better choice, so I give it  a 👍

makaroni4:
@jkowens any idea about failing AR 5 tests? Looks like latest AR RC has some issue with mysql, there is same error in other PR.

jkowens:
@makaroni4 yeah it turns out the seamless_database_pool gem is not currently compatible with ActiveRecord 5.0.0.rc1. I sent up a PR to fix, so hopefully there will be an updated release in the near future.

See bdurand/seamless_database_pool#25

cc: @zdennis

zdennis:
@makaroni4, seamless_database_pool is happy now in master thanks to @jkowens. Would you mind rebasing?

makaroni4:
@zdennis done, squashed commits into 1 🚀

zdennis:
Thank you @makaroni4 !

zdennis:

zdennis:
Bummer, this breaks [TravisCI](https://travis-ci.org/zdennis/activerecord-import/jobs/139376600), but I think that is a JDBCDriver error. I'm not going to hold up the release due to the error, will investigate later and will possibly skip if this is a known limitation with that driver.

Issue #94, opened 2013-05-03
inserted id


diogomaximo:
How can i get the inserted ids of my array object?
Can i do something like that:
ids = Entry.import entries

gudata:
Any solution on this?

jkowens:
@diogomaximo if using Postgres and activerecord-import 0.8.0+ you can do

```
result = Entry.import entries
ids = result.ids
```

Issue #230 would provide the same functionality to MySQL and SQLite once implemented.

PR #279, opened 2016-05-30
Return inserted ids MySQL

robertomiranda:
ref #94

### DIFF
--- a/lib/activerecord-import/adapters/mysql_adapter.rb
+++ b/lib/activerecord-import/adapters/mysql_adapter.rb
@@ -10,6 +10,8 @@ module ActiveRecord::Import::MysqlAdapter
   def insert_many( sql, values, *args ) # :nodoc:
     # the number of inserts default
     number_of_inserts = 0
+    first_inserted_id = nil
+    ids = []
 
     base_sql, post_sql = if sql.is_a?( String )
       [sql, '']
@@ -32,9 +34,9 @@ module ActiveRecord::Import::MysqlAdapter
 
     # if we can insert it all as one statement
     if NO_MAX_PACKET == max || total_bytes < max
-      number_of_inserts += 1
+      number_of_inserts = values.size
       sql2insert = base_sql + values.join( ',' ) + post_sql
-      insert( sql2insert, *args )
+      first_inserted_id = insert( sql2insert, *args )
     else
       value_sets = ::ActiveRecord::Import::ValueSetsBytesParser.parse(values,
         reserved_bytes: sql_size,
@@ -42,11 +44,16 @@ module ActiveRecord::Import::MysqlAdapter
       value_sets.each do |value_set|
         number_of_inserts += 1
         sql2insert = base_sql + value_set.join( ',' ) + post_sql
-        insert( sql2insert, *args )
+        ids << insert( sql2insert, *args )
       end
     end
 
-    [number_of_inserts, []]
+    if first_inserted_id
+      last_inserted_id = first_inserted_id + number_of_inserts - 1
+      ids = (first_inserted_id..last_inserted_id).to_a
+    end
+
+    [number_of_inserts, ids]
   end
 
   # Returns the maximum number of bytes that the server will allow


lib/activerecord-import/adapters/mysql_adapter.rb
```diff
@@ -32,21 +34,26 @@ def insert_many( sql, values, *args ) # :nodoc:
...
         sql2insert = base_sql + value_set.join( ',' ) + post_sql
-        insert( sql2insert, *args )
+        ids << insert( sql2insert, *args )
       end
     end
 
-    [number_of_inserts, []]
+    if first_inserted_id
+      last_inserted_id = first_inserted_id + number_of_inserts - 1
+      ids = (first_inserted_id..last_inserted_id).to_a
```
robertomiranda: @zdennis what do you think about this approach?

jkowens:
The database table would need to be InnoDB with `innodb_autoinc_lock_mode` set to 0 or 1 (default) to guarantee sequential id values for batch inserts. If it is set to 2 (interleaved) or is MyISAM, sequential ids cannot be guaranteed.

lib/activerecord-import/adapters/mysql_adapter.rb
```diff
@@ -32,21 +34,26 @@ def insert_many( sql, values, *args ) # :nodoc:
...
+      first_inserted_id = insert( sql2insert, *args )
     else
       value_sets = ::ActiveRecord::Import::ValueSetsBytesParser.parse(values,
         reserved_bytes: sql_size,
         max_bytes: max)
       value_sets.each do |value_set|
         number_of_inserts += 1
         sql2insert = base_sql + value_set.join( ',' ) + post_sql
-        insert( sql2insert, *args )
+        ids << insert( sql2insert, *args )
```
jkowens: This will result in missing ids because this is a batch operation.

lib/activerecord-import/adapters/mysql_adapter.rb
```diff
@@ -32,21 +34,26 @@ def insert_many( sql, values, *args ) # :nodoc:
 
     # if we can insert it all as one statement
     if NO_MAX_PACKET == max || total_bytes < max
-      number_of_inserts += 1
+      number_of_inserts = values.size
```
jkowens: This should be reverted to `number_of_inserts += 1` because it is not counting the number of rows inserted, but the actual number of insert operations executed.

lib/activerecord-import/adapters/mysql_adapter.rb
zdennis: @robertomiranda, thank you for contributing to ar-import! With regard to your question above I have the same concerns as @tonic20 and @jkowens. I'd want to ensure that we however this implemented we can guarantee reliability and avoid unsuspecting issues when the returned IDs are wrong.

philipgiuliani:
Would be nice to have this feature. It will work fine for all users with `INNODB + innodb_autoinc_lock_mode set to 0 or 1` Of course we'll have to mention that somewhere.

philipgiuliani:
This will also not work when there are duplicates and you use on_duplicate_key_update, right? Than it returns wrong id's probably.

jkowens:
@philipgiuliani you're correct, this would not work with on_duplicate_key_update or on_duplicate_key_ignore.

jkowens:
I'm going to close this PR for now because it doesn't seem like MySQL provides a good solution for this problem at the moment. With the need to support `ON DUPLICATE KEY UPDATE` and with UUID keys gaining momentum, I think we'll have to wait and hope MySQL eventually returns IDs like PostgreSQL.

shanecav84:
@jkowens Is it possible to get `recursive` to work with MariaDB (a popular community fork of MySQL). It supports:

- `INSERT...RETURNING`: https://mariadb.com/kb/en/insertreturning/
- `INSERT ON DUPLICATE KEY UPDATE`: https://mariadb.com/kb/en/insert-on-duplicate-key-update/

Issue #247, opened 2016-03-02
has_many associations don't save with recursive: :true


rhannequin:
Hello,

First, thank you very much for this gem, it is **much** appreciated every day from casual to complex data imports.

I am having troubles with the multi-level feature, when trying to import models with associations. I tried to follow the [multi-level example](https://github.com/zdennis/activerecord-import/wiki#multi-level-example) this way:

I created a blank app with two models:

```
$ bundle exec rails g model Book title:string
$ bundle exec rails g model Review name:string book:references
```

Then I added the association to the `Book` model:

```ruby
class Book < ActiveRecord::Base
  has_many :reviews
end
```

I created a basic rake task with the same code as the example, to insert books with one review in each, and checking at the end if everything went good:

```ruby
task test: :environment do
  Book.delete_all
  Review.delete_all

  books = []
  10.times do |i|
    book = Book.new(title: "Book ##{i}")
    book.reviews.build(name: "Review ##{i}")
    books << book
  end
  Book.import books, recursive: true

  puts "'Book' records count: #{Book.count}"
  puts "Last 'Book' reviews count: #{Book.last.reviews.count}"
  puts "'Review' records count: #{Review.count}"
end
```

What I have in return shows that the books were inserted but not the associated reviews:

```
'Book' records count: 10
Last 'Book' reviews count: 0
'Review' records count: 0
```

I don't know if my way of inserting associated models in database is wrong or if the multi-level feature has an issue, I thought it was the best place to warn about this behaviour has @sferik 's PR #230 (thank you for that btw) was supposed to fix it.

This test was run on Ubuntu 15.10 with:
* `rails` v4.2.5.1
* `activerecord` v4.2.5.1
* `sqlite3` v1.3.11
* `activerecord-import` from Github (at master@df866a4)

This post is a duplicate from my comment on #243.

Thank you.

JHFirestarter:
See @allenwlee findings in issue #231 (short answer is to add `autosave: true` and `inverse_of` in your models). 

2 problems I've found with this, in a common use case: 

[...]

PR #280, opened 2016-05-31
Upgrade seamless_database_pool to support AR 5.0

jkowens:
This will resolve the failing seamless_database_pool tests for AR 5.0. Note that AR 3.1 is not supported by the seamless_database_pool gem as of version 1.0.18 (which adds AR 5.0RC1 support).

PR #281, opened 2016-05-31
Add recursive import support for sqlite3

jkowens:
As requested in #247.

Sqlite doesn't support concurrent write operations, so it can be assumed all ids for a batch insert will be sequential.

This also extracts the recursive import tests out into a shared example so that they can be applied to other adapters. If all adapters eventually get this functionality, the specs could be moved into the main spec file.

### DIFF
--- a/test/support/shared_examples/recursive_import.rb
+++ b/test/support/shared_examples/recursive_import.rb
@@ -1,0 +1,122 @@
+def should_support_recursive_import
+  describe "importing objects with associations" do
+    let(:new_topics) { Build(num_topics, :topic_with_book) }
+    let(:new_topics_with_invalid_chapter) do
+      chapter = new_topics.first.books.first.chapters.first
+      chapter.title = nil
+      new_topics
+    end
+    let(:num_topics) { 3 }
+    let(:num_books) { 6 }
+    let(:num_chapters) { 18 }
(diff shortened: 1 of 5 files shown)


test/support/shared_examples/recursive_import.rb
```diff
@@ -0,0 +1,122 @@
...
+        assert_difference "Topic.count", +1 do
+          Topic.import new_topics, recursive: true, on_duplicate_key_update: [:updated_at], validate: false
+          new_topics.each do |topic|
+            assert_not_nil topic.id
+          end
+        end
+      end
+    end
+  end
+end
```
zdennis: :thumbsup: Thanks for pulling these out.

zdennis:
@jkowens, this looks good. It looks like AR 5.0 and AR 3.1 aren't happy. I'm less worried about 3.1 as we're going to deprecating that soon. 

Any thoughts on what's up with AR 5.0?

jkowens:
Hey @zdennis, merging #280 will fix the failing AR 5.0 tests. The latest version of Seamless Database Pool is now compatible with AR 5.0 RC1 but it is no longer compatible with AR 3.1.

zdennis:
Thanks @jkowens !

zdennis:

Issue #282, opened 2016-06-14
Polymorphic associations missing type column when imported via association


seanlinsley:
Initial support for polymorphic associations was added in #248 / #252, but given this:
```ruby
Book.last.discounts.import [:value], [[1]]
```
the `_type` column will not be included in the SQL

zdennis:
@seanlinsley, can you be more specific as to which column you're expecting to filled out that isn't? I think I know, but I'd rather not guess.

seanlinsley:
Sure. With a regular belongs_to association, you have a foreign key column including the name of the other model. From the tests, a chapter has a `book_id` column.

But for polymorphic associations, both the model name and the ID must be stored. From the tests, a discount has a `discountable_id` and a `discountable_type`.

My PR updates the code to also store `discountable_type`.

zdennis:
I'm sorry @seanlinsley I totally missed your PR when I typed my earlier comment. Thank you for being kind to me when I missed the obvious.

seanlinsley:
No worries at all. I should've been more clear when writing the ticket 😄

PR #283, opened 2016-06-14
Fix association import method for polymorphic associations

seanlinsley:
#282

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -61,7 +61,7 @@ class ActiveRecord::Associations::CollectionAssociation
 
       models.each do |m|
         m.public_send "#{symbolized_foreign_key}=", owner_primary_key_value
-        m.public_send "#{reflection.type}=", owner.class if reflection.type
+        m.public_send "#{reflection.type}=", owner.class.name if reflection.type
       end
 
       return model_klass.import column_names, models, options
@@ -85,7 +85,7 @@ class ActiveRecord::Associations::CollectionAssociation
 
       if reflection.type
         column_names << reflection.type
-        array_of_attributes.each { |attrs| attrs << owner.class }
+        array_of_attributes.each { |attrs| attrs << owner.class.name }
       end
 
       return model_klass.import column_names, array_of_attributes, options
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -61,6 +61,7 @@ class ActiveRecord::Associations::CollectionAssociation
 
       models.each do |m|
         m.public_send "#{symbolized_foreign_key}=", owner_primary_key_value
+        m.public_send "#{reflection.type}=", owner.class if reflection.type
       end
 
       return model_klass.import column_names, models, options
@@ -80,6 +81,11 @@ class ActiveRecord::Associations::CollectionAssociation
       else
         column_names << symbolized_foreign_key
         array_of_attributes.each { |attrs| attrs << owner_primary_key_value }
+      end
+
+      if reflection.type
+        column_names << reflection.type
+        array_of_attributes.each { |attrs| attrs << owner.class }
       end
 
       return model_klass.import column_names, array_of_attributes, options
--- a/test/import_test.rb
+++ b/test/import_test.rb
@@ -363,13 +363,13 @@ describe "#import" do
   end
 
   context "importing through an association scope" do
-    {has_many: :chapters, polymorphic: :discounts}.each do |association_type, association|
+    { has_many: :chapters, polymorphic: :discounts }.each do |association_type, association|
       let(:book)   { FactoryGirl.create :book }
       let(:scope)  { book.public_send association }
-      let(:klass)  { {chapters: Chapter, discounts: Discount}[association] }
-      let(:column) { {chapters: :title,  discounts: :amount }[association] }
-      let(:val1)   { {chapters: 'A',     discounts: 5       }[association] }
-      let(:val2)   { {chapters: 'B',     discounts: 6       }[association] }
+      let(:klass)  { { chapters: Chapter, discounts: Discount }[association] }
+      let(:column) { { chapters: :title,  discounts: :amount  }[association] }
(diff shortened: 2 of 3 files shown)


seanlinsley:
The tests pass. Instead it's a style violation:
```
Offenses:

test/import_test.rb:366:5: C: Space inside { missing.

    {has_many: :chapters, polymorphic: :discounts}.each do |association_type, association|

    ^

test/import_test.rb:366:50: C: Space inside } missing.

    {has_many: :chapters, polymorphic: :discounts}.each do |association_type, association|

                                                 ^

test/import_test.rb:369:22: C: Space inside { missing.

      let(:klass)  { {chapters: Chapter, discounts: Discount}[association] }

                     ^

test/import_test.rb:369:61: C: Space inside } missing.

      let(:klass)  { {chapters: Chapter, discounts: Discount}[association] }

                                                            ^

test/import_test.rb:370:22: C: Space inside { missing.

      let(:column) { {chapters: :title,  discounts: :amount }[association] }

                     ^

test/import_test.rb:371:22: C: Space inside { missing.

      let(:val1)   { {chapters: 'A',     discounts: 5       }[association] }

                     ^

test/import_test.rb:372:22: C: Space inside { missing.

      let(:val2)   { {chapters: 'B',     discounts: 6       }[association] }

                     ^
```

Is this really the preferred style? I've never seen a Ruby project with that preference.

zdennis:
> Is this really the preferred style? 

For now it is. We can always change it. 

>  I've never seen a Ruby project with that preference.

This may sound absurd, but I immediately imagined you pulling out a notepad, a pen, and a candle. After lighting the candle you carefully cross out a line item on a page titled "Bucket List":

> Bucket list:
>
> ~~[x] unheard of hash style preference~~
>

One can hope, right?

lib/activerecord-import/import.rb
```diff
@@ -61,6 +61,7 @@ def import(*args, &block)
 
       models.each do |m|
         m.public_send "#{symbolized_foreign_key}=", owner_primary_key_value
+        m.public_send "#{reflection.type}=", owner.class if reflection.type
```
zdennis: Should this be `owner.class.sti_name`?

test/import_test.rb
```diff
@@ -363,36 +363,30 @@
...
-    it "works importing array of columns and values" do
-      topic = FactoryGirl.create :topic
-      topic.books.import [:author_name, :title], [["Author #1", "Book #1"], ["Author #2", "Book #2"]]
-      assert_equal 2, topic.books.count
-      assert topic.books.detect { |b| b.title == "Book #1" && b.author_name == "Author #1" }
-      assert topic.books.detect { |b| b.title == "Book #2" && b.author_name == "Author #2" }
+        it "works importing array of columns and values" do
+          scope.import [column], [[val1], [val2]]
+
+          assert [val1, val2], scope.map(&column).sort
```
zdennis: I feel like the test has lost some readability in pulling everything up and out of the way.

lib/activerecord-import/import.rb
seanlinsley: sti for single table inheritance? that doesn't seem like the right method to call.

seanlinsley: Oh but `owner.class` wouldn't work for namespaced classes. I wonder what the correct method is to use, then.

test/import_test.rb
seanlinsley: Downside: every line of setup must be read to understand the tests themselves

Upside: it's very easy to add new tests without accidentally forgetting regular has_many, or forgetting polymorphic (because the each loop defines the test for each)

seanlinsley:
> For now it is. We can always change it. 

I'll update the code to use this style.

> Bucket list:

>~~[x] unheard of hash style preference~~

😆 well it's not unheard of in other languages (e.g. JS), I've just never seen a Ruby project prefer it.

seanlinsley:
Okay, looks like the tests pass on all but Rails 3.2:

```
Error: test_#import importing through an association scope for has_many_works importing array of columns and values(#import importing through an association scope for has_many):
  ArgumentError: assertion message must be String, Proc or Test::Unit::Assertions::AssertionMessage: <[]>(<Array>)

activerecord-import/test/import_test.rb:387:in `block (5 levels) in <top (required)>'
gems/activesupport-3.2.22.2/lib/active_support/testing/setup_and_teardown.rb:72:in `block in run'
gems/activesupport-3.2.22.2/lib/active_support/callbacks.rb:414:in `_run__635637341427027783__setup__4214765092007371939__callbacks'
gems/activesupport-3.2.22.2/lib/active_support/callbacks.rb:405:in `__run_callback'
gems/activesupport-3.2.22.2/lib/active_support/callbacks.rb:385:in `_run_setup_callbacks'
gems/activesupport-3.2.22.2/lib/active_support/callbacks.rb:81:in `run_callbacks'
gems/activesupport-3.2.22.2/lib/active_support/testing/setup_and_teardown.rb:70:in `run'

**Comment by seanlinsley** [comment]
Ah this is the problem: Rails 3.2 tries encoding the class as YAML:
```ruby
discountable_id: 1, discountable_type: "--- !ruby/class 'Book'\n"

File: lib/activerecord-import/import.rb (hunk 1 above)
Comment (seanlinsley) [inline lib/activerecord-import/import.rb]: This has been updated to do `owner.class.name`, which is the same thing that #252 does.

**Comment by seanlinsley** [comment]
The tests are green now.

File: lib/activerecord-import/import.rb [hunk 3]
```diff
@@ -85,7 +85,7 @@ def import(*args, &block)
 
       if reflection.type
         column_names << reflection.type
-        array_of_attributes.each { |attrs| attrs << owner.class }
+        array_of_attributes.each { |attrs| attrs << owner.class.name }
```
zdennis: :thumbsup:

zdennis:
Thanks @seanlinsley !

seanlinsley:
Thanks for being so responsive 😄

zdennis:

PR #284, opened 2016-06-15
Add the option to avoid serialization on import

YanivRan:
If you just want to load the text fields from string, there should be a way to avoid the serialization.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -67,7 +67,7 @@ class ActiveRecord::Associations::CollectionAssociation
 
     # supports empty array
     elsif args.last.is_a?( Array ) && args.last.empty?
-      return ActiveRecord::Import::Result.new([], 0, [])
+      return ActiveRecord::Import::Result.new([], 0, []) if args.last.empty?
 
     # supports 2-element array and array
     elsif args.size == 2 && args.first.is_a?( Array ) && args.last.is_a?( Array )
@@ -168,8 +168,7 @@ class ActiveRecord::Base
     # * +ignore+ - true|false, tells import to use MySQL's INSERT IGNORE
     #    to discard records that contain duplicate keys.
     # * +on_duplicate_key_ignore+ - true|false, tells import to use
-    #    Postgres 9.5+ ON CONFLICT DO NOTHING. Cannot be enabled on a
-    #    recursive import.
+    #    Postgres 9.5+ ON CONFLICT DO NOTHING.
     # * +on_duplicate_key_update+ - an Array or Hash, tells import to
     #    use MySQL's ON DUPLICATE KEY UPDATE or Postgres 9.5+ ON CONFLICT
     #    DO UPDATE ability. See On Duplicate Key Update below.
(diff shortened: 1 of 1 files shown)


zdennis:
@YanivRan , would you mind adding a test so I can better understand what this is achieving? Or could you post some code ? Thanks!

YanivRan:
Right now, the imported data for a text won't work if it's just a blurb of text and the field is a serialized field. I work with both your gem for loading and seed_dump to create the seeds. We have some serialized fields, that simply won't load in the serialized form. 
Seed_dump doesn't get the original data when creating a seed, so I added a pull request to the seed_dump to prevent serialization (get the original values), and if activerecord_import will also have this options, you won't need to worry about serialization in both creating and loading a seed file (you will just be able to load a string). This is added as an option switch, because it is optional, though probably beneficial in most cases.

YanivRan:
Will add a test soon, just have to find time to do so (at work).

lib/activerecord-import/import.rb
```diff
@@ -67,7 +67,7 @@ def import(*args, &block)
 
     # supports empty array
     elsif args.last.is_a?( Array ) && args.last.empty?
-      return ActiveRecord::Import::Result.new([], 0, [])
+      return ActiveRecord::Import::Result.new([], 0, []) if args.last.empty?
```
jkowens: I don't think this additional check is necessary?

lib/activerecord-import/import.rb
```diff
@@ -502,7 +501,7 @@ def import_without_validations_or_callbacks( column_names, array_of_attributes,
         end
       else
         values_sql.each do |values|
-          ids << connection.insert(insert_sql + values)
+          connection.execute(insert_sql + values)
```
jkowens: This block is for adapters that don't support bulk import, like old version of SQLite. I don't think it would need to change.

lib/activerecord-import/import.rb
YanivRan: these seems to been from the file itself. not something I added... it's from my forked file. so it's probably not needed.

lib/activerecord-import/import.rb
YanivRan: again, not something I added, but appeared in my fork. Weird, but not needed.

lib/activerecord-import/import.rb
jkowens: Yeah sorry, I edited my comment! 😉

lib/activerecord-import/import.rb
```diff
@@ -591,8 +590,8 @@ def values_sql_for_columns_and_attributes(columns, array_of_attributes) # :nodoc
               connection_memo.quote(type_caster.type_cast_for_database(column.name, val))
             elsif column.respond_to?(:type_cast_from_user)                                   # Rails 4.2 and higher
               connection_memo.quote(column.type_cast_from_user(val), column)
-            else                                                                             # Rails 3.2, 4.0 and 4.1
-              if serialized_attributes.include?(column.name)
+            else                                                                             # Rails 3.1, 3.2, 4.0 and 4.1
+              if serialized_attributes.include?(column.name) && (options[:serialized] != false )
```
YanivRan: This is the important part, avoid serialization if the serialized option is false. However, this only works for rails up to 4.1. If there's a way to make it work for all rails version it would be great.

sithara: @YanivRan  agree 👍 .. wish this would work for all rails versions :sob:

jkowens:
I'm closing this PR for now, please feel free to reopen with a cleaned up commit and further explanation of its purpose.

PR #287, opened 2016-06-22
Fix :no_returning option for JRuby

jkowens:
I'm not sure if this is the best possible solution, but I figure this will help shed light on the problem.


zdennis:
@jkowens, that was very very fast. :100: TY

zdennis:
I think this change is worth getting out in case someone is using jruby and wants to try out `no_returning: true`. I'm going to defer figuring out if it's the best to another day.  

I'm going to push out a 0.14.1 update with this in there.

Issue #286, opened 2016-06-18
Bulk Update


rjrobinson:
I'm curious as to if the import updates the same way that it imports.

I have run several tests where i have an array of models

I run `Model.import models` Then, run against the same and different data, and it seems to only initially create the record. If it is updating, the timestamps are not updating correctly.

seanlinsley:
It looks like [the tests](https://github.com/zdennis/activerecord-import/blob/8ba94f7c05b2259f6447aaf8b01ca5f4b0f5e0a2/test/support/postgresql/import_examples.rb#L96-L97) validate that `updated_at` is updated. Can you create a failing test for your scenario?

rjrobinson:
I believe that `:on_duplicate_key_update` might be an issue. I was not using that. Im curious. 

Do I have to match up the existing keys? (ie pull in the keys and when i Model.new(id: 1) ?) because `:name` is not unique in my case.

rjrobinson:
I guess what I am wondering is... 

if I have data, which are hosts. 

and I run `Host.import hosts`

This will almost always be an update call. 

```ruby
#<struct ActiveRecord::Import::Result
 failed_instances=
  [#<Host:0x007fe8bdac98a8
    id: 1,
    address: "127.0.0.1",
    checks_enabled: true,
    display_name: nil,
    contact_groups: nil,
    last_check: "1448915437",
    last_notification: "0",
    last_state: "0",
    last_state_change: "1384209316",
    last_time_down: "0",
    last_time_unreachable: "0",
    last_time_up: "1448915437",
    name: "RJs Home Server",
    next_check: "1448915737",
    notification_interval: 60,
    notifications_enabled: true,
    num_services: 1,
    num_services_crit: 0,
    num_services_hard_crit: 0,
    num_services_hard_ok: 1,
    num_services_hard_unknown: 0,
    num_services_hard_warn: 0,
    num_services_ok: 1,
    num_services_pending: 0,
    num_services_unknown: 0,
    num_services_warn: 0,
    should_be_scheduled: true,
    state: 0,
    state_type: 1,
    total_services: 1,
    created_at: Sat, 18 Jun 2016 17:31:51 EDT -04:00,
    updated_at: Sat, 18 Jun 2016 17:31:51 EDT -04:00,
    last_known_source: "op5.sys.comcast.net",
    snow_sys_id: nil,
    snow_successful: nil,
    snow_updated_at: nil>],
 num_inserts=0,
 ids=[]>
```

[...]

PR #290, opened 2016-06-23
Raise ArgumentError if no conflict_target available

jkowens:
If no conflict target is specified, and a table does not have a primary key, an `ArgumentError` is raised with the message:

```
Expected :conflict_target or :constraint_name to be specified
```

Currently, a `ActiveRecord::StatementInvalid` error is raised because the SQL is generated without any conflict target value:

```
ON CONFLICT () DO UPDATE SET
```

### DIFF
--- a/test/support/postgresql/import_examples.rb
+++ b/test/support/postgresql/import_examples.rb
@@ -170,24 +170,53 @@ def should_support_postgresql_upsert_functionality
           should_update_fields_mentioned
         end
 
-        context "with no :conflict_target or :constraint_name" do
+        context "default to the primary key" do
           let(:columns) { %w( id title author_name author_email_address parent_id ) }
           let(:values) { [[100, "Book", "John Doe", "john@doe.com", 17]] }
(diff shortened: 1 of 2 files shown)


test/support/postgresql/import_examples.rb
```diff
@@ -170,25 +170,51 @@ def should_support_postgresql_upsert_functionality
...
+            should_support_on_duplicate_key_update
+            should_update_fields_mentioned
+          end
+        end
+
+        context "with no primary key" do
+          it "raises ArgumentError" do
+            assert_raises ArgumentError do
+              Widget.import Build(1, :widgets), on_duplicate_key_update: [:data], validate: false
+            end
```
seanlinsley: I feel like it would be better to expect the explanatory error message, instead of simply expecting an `ArgumentError`

seanlinsley:
Ref: https://github.com/zdennis/activerecord-import/issues/286#issuecomment-227925835

test/support/postgresql/import_examples.rb
jkowens: I updated the test, let me know if it is inline with what you are thinking. Thanks!

zdennis:
Thanks @jkowens

PR #299, opened 2016-07-08
Improve test coverage for serailzed attributes

mrloop:
While trying to track down bug between activerecord-import and another gem wrote more test coverage to rule out that issue was caused by activerecord-import. This pull requests only adds more test coverage.


zdennis:
Thank you @mrloop !

Issue #228, opened 2016-01-26
Bulk uniqueness validation


sferik:
If I bulk import _n_ records with a uniqueness constraint, _n_ `SELECT` queries will be executed before the `INSERT`.

For example, given the following Active Record model:

``` ruby
class User < ActiveRecord::Base
  validates :email, uniqueness: true
end
```

``` ruby
> users = 10.times.map { User.new(email: "#{rand}@example.com") }
> User.import(users)
  User Exists (2.1ms)  SELECT  1 AS one FROM "users" WHERE "users"."email" = '0.5941205541221567@example.com' LIMIT 1
  User Exists (0.5ms)  SELECT  1 AS one FROM "users" WHERE "users"."email" = '0.7343287052445423@example.com' LIMIT 1
  User Exists (0.7ms)  SELECT  1 AS one FROM "users" WHERE "users"."email" = '0.322860914726271@example.com' LIMIT 1
  User Exists (0.3ms)  SELECT  1 AS one FROM "users" WHERE "users"."email" = '0.4479100997353238@example.com' LIMIT 1
  User Exists (0.4ms)  SELECT  1 AS one FROM "users" WHERE "users"."email" = '0.8518056478659825@example.com' LIMIT 1
  User Exists (0.2ms)  SELECT  1 AS one FROM "users" WHERE "users"."email" = '0.12517651279963693@example.com' LIMIT 1
  User Exists (0.2ms)  SELECT  1 AS one FROM "users" WHERE "users"."email" = '0.8874450255741644@example.com' LIMIT 1
  User Exists (0.2ms)  SELECT  1 AS one FROM "users" WHERE "users"."email" = '0.5142328998773664@example.com' LIMIT 1
  User Exists (0.3ms)  SELECT  1 AS one FROM "users" WHERE "users"."email" = '0.863622688265132@example.com' LIMIT 1
  User Exists (0.2ms)  SELECT  1 AS one FROM "users" WHERE "users"."email" = '0.4563742277940873@example.com' LIMIT 1
```

Ideally, these 10 queries would be combined into a single `SELECT`:

``` sql
[...]

PR #301, opened 2016-07-08
Bypass uniqueness validation on import

jkowens:
Improves performance of importing with validations when models have uniqueness validators (See #228). Uniqueness cannot be guaranteed if there are duplicate values in the batch. Race conditions are also possible, so the only way to really preserve uniqueness is by using a database index.

With uniqueness validation removed, model validation could be done in combination with `:on_duplicate_key_update`, `:on_duplicate_key_ignore`, or `:ignore` for databases that support those options.

Just looking for feedback at this point, thanks!

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -22,6 +22,44 @@ module ActiveRecord::Import #:nodoc:
       super "Missing column for value <#{name}> at index #{index}"
     end
   end
+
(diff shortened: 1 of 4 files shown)


zdennis:
I added a "wip" label just now, just FYI that one exists from here on out for wip-y things.

ioquatix:
Good idea.

PR #309, opened 2016-08-24
Same primary key on recursive issue resolved

chopraanmol1:
While using recursive option same primary key is used for all activerecords. Which fails when a activerecord has different primary key name or doesn't have one.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -332,8 +332,9 @@ class ActiveRecord::Base
     end
 
     def import_helper( *args )
-      options = { validate: true, timestamps: true, primary_key: primary_key }
+      options = { validate: true, timestamps: true }
       options.merge!( args.pop ) if args.last.is_a? Hash
+      options[:primary_key] = primary_key
 
       # Don't modify incoming arguments
       if options[:on_duplicate_key_update]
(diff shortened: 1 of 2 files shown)


jkowens:
@chopraanmol1 this is good. If could make sure this has test coverage that would really help get this merged in. I think you could just change the `end_notes` table in `generic_schema.rb` to use a primary key of `end_note_id` and that would do it since there is a recursive import test that imports end notes along with chapters.

lib/activerecord-import/import.rb
```diff
@@ -332,8 +332,9 @@ def import!(*args)
     end
 
     def import_helper( *args )
-      options = { validate: true, timestamps: true, primary_key: primary_key }
+      options = { validate: true, timestamps: true }
       options.merge!( args.pop ) if args.last.is_a? Hash
+      options[:primary_key] = primary_key
```
jkowens: A comment might be a good idea here so that this doesn't get lost down the road.

chopraanmol1:
@jkowens does this looks good now

zdennis:
@chopraanmol1 , I just pulled this down and your changes are great! Thank you for submitting the fix.

Thank you @jkowens for shepherding this thru!

jkowens:
This has be included in the 0.16.0 release. Thanks @chopraanmol1!

Issue #138, opened 2014-04-23
#import raises PG::NotNullViolation for id field on PostgreSQL


cspeer:
I recently migrated from MySQL to PostgreSQL and now I'm getting this when using #import:
```
ActiveRecord::StatementInvalid: PG::NotNullViolation: ERROR:  null value in column "id" violates not-null constraint
```

The INSERT-Statement it produces clearly lists id as one of the fields and a NULL value is provided:
```
INSERT INTO "metrics_user_top_list_most_active" ("id","object_id","amount","at","created_at","updated_at","customer_id") VALUES (NULL, ...
```

I'm using activerecord-import 0.5.0 and Rails 4.0.4 so I guess for the id field to be NOT NULL is quite common. Any ideas what's going wrong here?

Thanks,
Chris

avishai:
Any news on this one?

tokhi:
Any updates on this issue? I have the same problem.

jkowens:
@tokhi as @avishai mentioned, it sounds like this is an issue with the MySQL auto-increment primary key column not being properly setup with a sequence in the Postgres database on migration. Can you you try @avishai's queries and report back?

It looks like [PGLoader](https://github.com/dimitri/pgloader) attempts to help with migrations to Postgres.

lefty313:
@avishai Thank your sir it saved my day :+1:

coreyward:
I have this issue with Postgresql as well on a database a table like this:

```sql
CREATE TABLE answers (
    id uuid DEFAULT uuid_generate_v4() NOT NULL,
    answer character varying,
    created_at timestamp without time zone NOT NULL,
    updated_at timestamp without time zone NOT NULL
);
```
When AR generates SQL for these records, it simply doesn't specify the `id` column at all. For example: 

```sql
INSERT INTO "answers" ("answer", "created_at", "updated_at") VALUES ($1, $2, $3, $4, $5, $6) RETURNING "id"  [["answer", "Yes"], ["created_at", "2016-06-16 04:18:20.508992"], ["updated_at", "2016-06-16 04:18:20.508992"]]
```

Whereas `import` generates the following (abbreviated):

```sql
INSERT INTO "answers" ("id", "answer", "created_at", "updated_at") VALUES (NULL, 'Yes', '2016-06-16 04:14:21.168674', '2016-06-16 04:14:21.168791')
```

steventen:
I just found the same problem as @coreyward  mentioned. 

[...]

PR #312, opened 2016-08-26
Support uuid primary keys

### LINKED ISSUE #311: Support non-integer primary keys
I'm using UUID's are my primary key. Whenever I run an import my id's disappear. I believe this is because you are expecting the primary key to be an integer.

I believe this is the line that's causing the issue:

https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/import.rb#L534

jkowens:
Resolves #311 and the issue that #138 has evolved into.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -347,10 +347,14 @@ class ActiveRecord::Base
       if args.last.is_a?( Array ) && args.last.first.is_a?(ActiveRecord::Base)
         if args.length == 2
           models = args.last
-          column_names = args.first
+          column_names = args.first.dup
         else
           models = args.first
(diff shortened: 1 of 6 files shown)


steventen:
The fix looks good to me. I've run my previous failed tests against this branch and they pass.

Thanks!

anthonator:
👍 looks good to me!

jkowens:
@zdennis do you think this could be merged into master? I have another update that could make use of some of this work. Trying to take advantage of Hacktoberfest! 😄

zdennis:
@jkowens, I added you as a collaborator. Want to give it a go on this PR to make sure I pushed the right button? :)

jkowens:
Looks like I'm good to go! 👍

PR #314, opened 2016-09-06
Updating support for import from a STI subclass with polymorphic associations 

JNajera:
The commits fixes the import a subclass of STI with polymorphic associations. It was getting the model class name to the polymorphic record instead of its parent class name.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -582,7 +582,8 @@ class ActiveRecord::Base
           child.public_send("#{association_reflection.foreign_key}=", model.id)
           # For polymorphic associations
           association_reflection.type.try do |type|
-            model_class_name = model.class.superclass.name == "ActiveRecord::Base" ? model.class.name : model.class.superclass.name
+            model_class_name = model.class.superclass.name
+            model_class_name = model.class.name if ["ActiveRecord::Base", "ApplicationRecord"].include? model_class_name
             child.public_send("#{type}=", model_class_name)
           end
         end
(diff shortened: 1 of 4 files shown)


lib/activerecord-import/import.rb
```diff
@@ -582,7 +582,8 @@ def find_associated_objects_for_import(associated_objects_by_class, model)
           child.public_send("#{association_reflection.foreign_key}=", model.id)
           # For polymorphic associations
           association_reflection.type.try do |type|
-            child.public_send("#{type}=", model.class.name)
+            model_class_name = model.class.superclass.name == "ActiveRecord::Base" ? model.class.name : model.class.superclass.name
```
jkowens: In Rails 5 ActiveRecord objects extend `ApplicationRecord` instead of `ActiveRecord::Base`.

JNajera: Thanks for the comment. I will get it fixed.

hoffmanilya:
Amazing! I just ran into this problem and you posted a fix an hour ago...

Thanks - works like a charm.

lib/activerecord-import/import.rb
```diff
@@ -582,7 +582,9 @@ def find_associated_objects_for_import(associated_objects_by_class, model)
           child.public_send("#{association_reflection.foreign_key}=", model.id)
           # For polymorphic associations
           association_reflection.type.try do |type|
-            child.public_send("#{type}=", model.class.name)
+            model_class_name = model.class.superclass.name
+            model_class_name = model.class.name if ["ActiveRecord::Base", "ApplicationRecord"].include? model_class_name
```
jkowens: I think this can be simplified by just calling:

`child.public_send("#{type}=", model.base_class.name)`

JNajera: Right. Its cleaner.

jkowens:
Cool, good catch!

jkowens:
This has be included in the 0.16.0 release. Thanks @JNajera!

PR #315, opened 2016-09-12
fix setting returned ids to wrong models

mizukami234:
In `support_setting_primary_key_of_imported_objects? == true` adapters like Postgres, `set_ids_and_mark_clean` iterates returned ids and set them to models (include both valid and invalid). It seems that this method should ignore invalid models.

### DIFF
--- a/lib/activerecord-import/adapters/sqlite3_adapter.rb
+++ b/lib/activerecord-import/adapters/sqlite3_adapter.rb
@@ -33,8 +33,8 @@ module ActiveRecord::Import::SQLite3Adapter
     value_sets.each do |value_set|
       number_of_inserts += 1
       sql2insert = base_sql + value_set.join( ',' ) + post_sql
-      first_insert_id = insert( sql2insert, *args )
-      last_insert_id = first_insert_id + value_set.size - 1
+      last_insert_id = insert( sql2insert, *args )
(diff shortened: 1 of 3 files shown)


jkowens:
@mizukami234 nice! 💯 

 It looks like you just have a couple rubocop issues to fix, and the tests should pass.

lib/activerecord-import/adapters/sqlite3_adapter.rb
```diff
@@ -33,8 +33,8 @@ def insert_many(sql, values, *args) # :nodoc:
     value_sets.each do |value_set|
       number_of_inserts += 1
       sql2insert = base_sql + value_set.join( ',' ) + post_sql
-      first_insert_id = insert( sql2insert, *args )
-      last_insert_id = first_insert_id + value_set.size - 1
+      last_insert_id = insert( sql2insert, *args )
+      first_insert_id = last_insert_id - value_set.size + 1
```
mizukami234: It looks like `insert` returns last inserted id, not first:
https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb#L126

mizukami234:
Thanks @jkowens 
Now I've fixed them all ✅

jkowens:
This has be included in the 0.16.0 release. Thanks @mizukami234!

Issue #186, opened 2015-05-04
Import + hstore


pkurek:
Hello,

Is it possible to use `AR.import` with postgresql hstore extension ? Right now I'm getting this error

```ruby
ActiveRecord::Import::MissingColumnError: Missing column for value <field> at index 0
```
Edit 1:

I think the gem itself has problems with STI in general. Example it duplicates the "type" field when trying to inject.

jkowens:
I'm not sure about STI, but the MissingColumnError could be caused by the way the store values are being set. See #150 and #151. Could you post sample code of how you are attempting to import?

PR #322, opened 2016-10-06
Add store accessor support for json, jsonb, and hstore datatypes

### LINKED ISSUE #313: JSONB support
There's an error on [import.rb #360](https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/import.rb#L360). It always returns an empty hash ({}) for JSONB columns, so no value is saved on the database!

```ruby
model.read_attribute_before_type_cast(name.to_s)
```

What is the problem of using _**model[name.to_s]**_ instead? It seems it works for me. I'm guessing for some other types it wouldn't.

Any workaround? Any plan to update the gem to support JSONB columns?

jkowens:
Add support for setting attributes on serialized fields using store accessor. JSON, JSONB and HSTORE data types require AR 4.0 and up. Stores backed by a text field are supported in Rails 3.2.

Resolves #313, #186 and #150.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,10 +27,13 @@ matrix:
 before_script:
   - mysql -e 'create database activerecord_import_test;'
   - psql -c 'create database activerecord_import_test;' -U postgres
-  - psql -U postgres -c "create extension postgis"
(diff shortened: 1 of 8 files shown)


jkowens:
@mparodi and @anthonator could you confirm if this update resolves your issue?

PR #332, opened 2016-10-11
prevent removing errors messages by duping original models error messages to failed instances

### LINKED ISSUE #331: not possible to read errors messages since 0.14.0 on failed instances
Hello, 
method `import_with_validations( column_names, array_of_attributes, options = {} )`

related commit  https://github.com/zdennis/activerecord-import/commit/d0322abe5d40c69fd89672f56e32bfa10fc8c150

before this commit implementation returned validated models

```
 failed_instances << instance
```
after
```
 failed_instances << model.dup
```
This makes it impossible to read error messages 
Example

```
model.errors.full_messages
=> ["Last name has already been taken"]

model.dup.errors.full_messages
=> []
```

Fivell:
fixes #331

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -476,7 +476,9 @@ class ActiveRecord::Base
           next if model.valid?(options[:validate_with_context])
           raise(ActiveRecord::RecordInvalid, model) if options[:raise_error]
           array_of_attributes[i] = nil
-          failed_instances << model.dup
+          failure = model.dup
+          failure.errors.send(:initialize_dup, model.errors)
+          failed_instances << failure
         end
       end
 
--- a/test/import_test.rb
+++ b/test/import_test.rb
@@ -121,7 +121,10 @@ describe "#import" do
       it "should report the failed instances" do
         results = Topic.import columns, invalid_values, validate: true
         assert_equal invalid_values.size, results.failed_instances.size
-        results.failed_instances.each { |e| assert_kind_of Topic, e }
+        results.failed_instances.each do |e|
+          assert_kind_of Topic, e
+          assert_equal e.errors.count, 1
+        end
       end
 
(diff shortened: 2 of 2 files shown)


lib/activerecord-import/import.rb
```diff
@@ -476,7 +476,7 @@ def import_with_validations( column_names, array_of_attributes, options = {} )
           next if model.valid?(options[:validate_with_context])
           raise(ActiveRecord::RecordInvalid, model) if options[:raise_error]
           array_of_attributes[i] = nil
-          failed_instances << model.dup
```
jkowens: The reason the model is duped is because it is being reused to validate attribute values instead of instantiating a new instance for every record. I think it might be possible to dupe the errors as well.

Can you try something like:
```
failure = model.dup
failure.errors.send(:initialize_dup, model.errors)
```

jkowens:
Thanks for reporting this issue, it definitely needs fixed!

lib/activerecord-import/import.rb
jkowens: See https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/import.rb#L473. The variable `model` is instantiated outside of the loop so it is being reused to save the time of repeatedly instantiating a new one.

If you don't dupe it, `failed_instances` will be an array of the same object.

Fivell: >>Can you try something like:

done

jkowens:
Sorry, one last thing, could you amend the commit message?

Fivell:
@jkowens , sure

jkowens:
Thanks again!

Fivell:
@jkowens , you're welcome!
BTW any thoughts time of release? I need it to close another issue in dependent project :)

jkowens:
This was just released in v0.16.1 🚀

PR #335, opened 2016-10-15
Update version constraint for rails 3.2 in readme

cydparser:
The original wording implied that a constraint of `~> 0.3.0` should be used, but newest version works fine with rails 3.2.


jkowens:
Thanks for this update!

PR #337, opened 2016-11-09
Avoid accessing ActiveRecord::Base.connection directly for multiple databases

KentoMoriwaki:
When using multiple databases by extending ActiveRecord::Base, `ActiveRecord::Base.connection` refers wrong connection.

Therefore simply use `self.query_cache` instead of `ActiveRecord::Base.connection.query_cache`.

### DIFF
--- a/lib/activerecord-import/adapters/postgresql_adapter.rb
+++ b/lib/activerecord-import/adapters/postgresql_adapter.rb
@@ -21,7 +21,7 @@ module ActiveRecord::Import::PostgreSQLAdapter
       insert( sql2insert, *args )
     end
 
-    ActiveRecord::Base.connection.query_cache.clear
+    query_cache.clear
 
     [number_of_inserts, ids]
   end


lib/activerecord-import/adapters/postgresql_adapter.rb
```diff
@@ -21,7 +21,7 @@ def insert_many( sql, values, *args ) # :nodoc:
       insert( sql2insert, *args )
     end
 
-    ActiveRecord::Base.connection.query_cache.clear
+    query_cache.clear
```
jkowens: @KentoMoriwaki could you explain why this line of code is needed? I'm trying to make sure there is test coverage. I found this test, but still don't really understand the purpose:

https://github.com/zdennis/activerecord-import/blob/master/test/support/postgresql/import_examples.rb#L23

Thanks!

jkowens: Yes, I think that would be much better! Could you also take a look at the test and make sure it's actually testing what we want. It doesn't look to me like the existing test is useful.

jkowens:
@KentoMoriwaki will you have time to update this commit to use an uncached query?

jkowens:
I'm going to merge this as is because I did some testing and using `uncached` for the insert query results in broken tests. This might be worth further investigation later.

KentoMoriwaki:
I'm sorry for my late reply. Thanks for merging!

jkowens:
@KentoMoriwaki this is fixed in version 0.16.2.

PR #342, opened 2016-11-15
Allow uuid primary keys to be manually set

### LINKED ISSUE #341: Import not respecting uuid primary keys
Greetings,

It looks like ActiveRecord::Import#import_helper was changed to remove columns from the array of columns to populate if it is a primary key and a uuid (https://github.com/zdennis/activerecord-import/blame/master/lib/activerecord-import/import.rb#L370).

Digging around it looks like this code was added as a result of https://github.com/zdennis/activerecord-import/issues/138

One side effect of this is that although it fixes the not-null exception in the mentioned issue, it now prevents callers from explicitly setting the primary key's value if it is a UUID.

So I'm wondering if we want to tighten this up somehow and accommodate both cases. If not, is there another way I can accomplish this?

jkowens:
Fixes #341.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -367,7 +367,7 @@ class ActiveRecord::Base
           column_names = self.column_names.dup
         end
 
-        if column_names.include?(primary_key) && columns_hash[primary_key].type == :uuid
+        if models.first.id.nil? && column_names.include?(primary_key) && columns_hash[primary_key].type == :uuid
           column_names.delete(primary_key)
         end
 
--- a/test/support/postgresql/import_examples.rb
+++ b/test/support/postgresql/import_examples.rb
@@ -59,19 +59,34 @@ def should_support_postgresql_import_functionality
     end
   end
 
-  describe "with a uuid primary key" do
-    let(:vendor) { Vendor.new(name: "foo") }
-    let(:vendors) { [vendor] }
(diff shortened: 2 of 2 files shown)


lib/activerecord-import/import.rb
```diff
@@ -367,7 +367,7 @@ def import_helper( *args )
           column_names = self.column_names.dup
         end
 
-        if column_names.include?(primary_key) && columns_hash[primary_key].type == :uuid
+        if models.first.id.nil? && column_names.include?(primary_key) && columns_hash[primary_key].type == :uuid
```
jkowens: The `id` method in on an ActiveRecord model always returns the primary key whether it is named 'id' or something else.

I'm not sure specifying some ID's but not all would work. You would end up attempting to insert NULL ID values which would result in a SQL error.

Dclusin-og: Ok is fine with me then

jkowens: Hm maybe an error is a better result, since that should be an exceptional scenario if some IDs have values and some do not.

Dclusin-og:
Awesome thanks a lot @jkowens. Would it be possible to push this up to a point release on ruby gems?

jkowens:
@Dclusin-og will do, just trying to get one more PR ready to go.

jkowens:
@Dclusin-og this is available in version 0.16.2.

Issue #54, opened 2012-05-10
Support REPLACE and INSERT IGNORE


avit:
The `:on_duplicate_key_update` option is fine when only parts of existing records need to change. It would be helpful to also support the native `REPLACE` syntax directly when the whole record needs to be replaced, and `INSERT IGNORE` when existing records should remain untouched.

Ideas for syntax?

`:on_duplicate_key => :ignore`

`:on_duplicate_key => :replace`

octavpo:
I just found about the :ignore option, which is nice, but it still has an issue: when set to true the code still inserts an 'ON DUPLICATE KEY UPDATE' clause for the 'updated_at' field. I think it's clear it shouldn't. Not even sure which one takes precedence.

I see also that there's another option :on_duplicate_key_ignore for Postgres. I don't know Postgres, but isn't that the same behavior as the :ignore key for MySQL? If so it would be good if they used the same key.

Thanks.

jkowens:
REPLACE is potentially dangerous and with ON DUPLICATE KEY support, I don't think there is a reason to introduce that functionality.

@octavpo :ignore does not apply to Postgres, it is MySQL only. For Postgres use `:on_duplicate_key_ignore`. Ideally `:on_duplicate_key_ignore` will one day be an alias for `:ignore` for MySQL.

jkowens:
@octavpo the issue you mentioned with 'ON DUPLICATE KEY UPDATE' being included when :ignore option is used for MySQL will be fixed in the next release. It will also allow :ignore or :on_duplicate_key_ignore to be used interchangeably for MySQL, Postgres, and SQLite.

PR #348, opened 2016-12-15
Add MySQL REPLACE support

caulfield:
Hi.

This PR adds mysql REPLACE functionallity. 
[Docs](http://dev.mysql.com/doc/refman/5.7/en/replace.html) from mysql site.
I'm not very experienced with mysql and this gem, let me know if I missed something.

Up #54

### DIFF
--- a/test/support/shared_examples/replace_import.rb
+++ b/test/support/shared_examples/replace_import.rb
@@ -1,0 +1,40 @@
+def should_support_replace
+  describe '#supports_replace?' do
+    it "returns true" do
+      assert ActiveRecord::Base.supports_replace?
+    end
+  end
+
+  describe '#import' do
+    let(:columns) { %w( id shopping_cart_id book_id ) }
+    let(:values) { [[1, 1, 2], [2, 1, 3]] }
+    let(:new_values) { [[3, 1, 2], [4, 1, 4]] }
+
+    macro(:perform_import) do |*opts|
+      CartItem.import columns, new_values, opts.extract_options!.merge(replace: true)
+    end
+
+    setup do
+      CartItem.import columns, values
+    end
+
+    it "should replace conflicting rows" do
+      perform_import
+      refute CartItem.find_by(id: 1).present?, "deletes conflicting rows"
+      assert CartItem.find_by(id: 3).present?, "inserts new rows"
+    end
+
+    it "should insert new rows" do
+      perform_import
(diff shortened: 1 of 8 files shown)


caulfield:
It's a very strange CI error. Do you have any thoughts? Everything is good locally

jkowens:
@caulfield have you run into a situation were you needed REPLACE instead of UPSERT functionality?

caulfield:
@jkowens, I created this pull request based on my current project requirements. If your database has a strict schema and you have no ability to change keys(primary). REPLACE helped me to keep consistency without running ruby code, it matters for performance and bulk inserts.
In my opinion, gem should provide an interface for communication with a database. Side effects(as an example of dangerous things) should be resolved by developers.
I understand that it's not a very popular feature, I can add more docs in README with links to MySQL docs.

jkowens:
I still don't understand why ON DUPLICATE KEY UPDATE could not be used instead? Could you provide me with more detail about the reasons REPLACE would have to be used? 

Each feature that is added to a gem adds more code to support long term. If there isn't a strong reason to add a feature (or there is an existing alternative) then I think that is good reason to give pause and ask if it is really necessary.

I appreciate all the work that you've put into this patch, it is very thorough. If you can help me see a scenario where ON DUPLICATE KEY UPDATE could not be used in place of REPLACE, that may help sway my thinking on this addition.

caulfield:
It makes sense with more code support.
The original problem was about inserting a lot of records without checking unique indices' values before. `REPLACE` gave me the more convenient way to insert new records. I receive a lot of time series data from 3rd party sources and I have to update(delete and then create new with default values) rows with duplicated keys. There weren't ids of records from this source, so I used a composite unique index.
I understand that `REPLACE` is less popular with `ON DUPLICATE KEY UPDATE` and the same task could be solved by both methods. The merge decision is up to you.

test/support/shared_examples/replace_import.rb
```diff
@@ -0,0 +1,40 @@
...
+      CartItem.import columns, new_values, opts.extract_options!.merge(replace: true)
+    end
+
+    setup do
+      CartItem.import columns, values
+    end
+
+    it "should replace conflicting rows" do
+      perform_import
+      refute CartItem.find_by(id: 1).present?, "deletes conflicting rows"
```
jkowens: I think the use of `#find_by` is causing tests for AR 3.2 to fail.

jkowens:
Closing for now.

PR #371, opened 2017-01-24
Properly set returned ids when using composite primary keys

### LINKED ISSUE #369: Model primary key is set to 0 after import with composite_primary_keys
````ruby
class Tag < ActiveRecord::Base
  self.primary_keys = :tag_id, :publisher_id
end

tags = [Tag.new(tag_id: 1, publisher_id: 1, tag: 'Mystery')]
Tag.import tags
tags.first

=> #<Tag:0x0055b6b1e5b010
 tag_id: 0,
 publisher_id: 1,
 tag: 'Mystery'>
````

Note that `tag_id` was set to `0`. In database the value was saved correctly.

I've investigated the cause, but don't know how to proper fix it.

The problem is that `ActiveRecord::Import::PostgreSQLAdapter#insert_many` is returning `ids` as `"(1,1)"`.

Then in `#set_ids_and_mark_clean` the line `model.id = id` will execute https://github.com/composite-primary-keys/composite_primary_keys/blob/master/lib/composite_primary_keys/base.rb#L106
that will call `CompositePrimaryKeys::CompositeKeys#parse` 
which expect id to be `"1,1"` (without parenthesis) https://github.com/composite-primary-keys/composite_primary_keys/blob/master/lib/composite_primary_keys/composite_arrays.rb#L30. So with `"(1,1)"` it will parse as `["(1", "1)"]`, which will be converted to `0` and `1`.

### PR DESCRIPTION
**Pull request by jkowens**
Resolves #369.

### DIFF
--- a/lib/activerecord-import/adapters/postgresql_adapter.rb
+++ b/lib/activerecord-import/adapters/postgresql_adapter.rb
@@ -4,7 +4,8 @@ module ActiveRecord::Import::PostgreSQLAdapter
 
   MIN_VERSION_FOR_UPSERT = 90_500
 
-  def insert_many( sql, values, *args ) # :nodoc:
+  def insert_many( sql, values, options = {}, *args ) # :nodoc:
+    primary_key = options[:primary_key]
     number_of_inserts = 1
     ids = []
 
@@ -15,11 +16,17 @@ module ActiveRecord::Import::PostgreSQLAdapter
     end
 
     sql2insert = base_sql + values.join( ',' ) + post_sql
-    if post_sql =~ /RETURNING\s/
-      ids = select_values( sql2insert, *args )
-      query_cache.clear if query_cache_enabled
-    else
+
+    if primary_key.blank? || options[:no_returning]
       insert( sql2insert, *args )
+    else
+      ids = if primary_key.is_a?( Array )
+        # Select composite primary keys
+        select_rows( sql2insert, *args )
+      else
+        select_values( sql2insert, *args )
+      end
+      query_cache.clear if query_cache_enabled
     end
 
     [number_of_inserts, ids]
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: lib/activerecord-import/import.rb [hunk 1]
```diff
@@ -598,6 +598,7 @@ def set_attributes_and_mark_clean(models, import_result, timestamps)
       return if models.nil?
       models -= import_result.failed_instances
       import_result.ids.each_with_index do |id, index|
+        id.tr!('()', '') if id.is_a? String
```
guigs: Hum, now the format of result.ids are different whether you import as models or arrays (because set_attributes_and_mark_clean is only called for models and because `tr!` modifies the object).

Also it would cause problems if someone is using string as primary key and they happen to contain `(` or `)`. Unlikely, but possible.

Could we parse ids in database adapter, and return an array of arrays, like `[[1, 2], [3, 4]]`?

jkowens: [[1, 2], [3, 4]] -- that is what I was expecting to be returned. If there is a way to make that happen, that would definitely be ideal!

lib/activerecord-import/adapters/postgresql_adapter.rb
```diff
@@ -15,11 +16,17 @@ def insert_many( sql, values, *args ) # :nodoc:
...
-      ids = select_values( sql2insert, *args )
-      query_cache.clear if query_cache_enabled
-    else
+
+    if primary_key.blank? || options[:no_returning]
       insert( sql2insert, *args )
+    else
+      ids = if primary_key.is_a?( Array )
+        # Select composite primary keys
+        select_rows( sql2insert, *args )
```
jkowens: @guigs so I found select_rows is the method needed for the job 🎉

PR #379, opened 2017-02-07
Fixed missing quotes when returnings ids in postgresql

hugobgranja:
Missing quotes cause an error where postgresql can't recognize the columns if their name is not entirely in lowercase letters.


jkowens:
@hugobgranja thanks for providing a fix! Could you change one of the column names in our test schema to make the tests fail first (and add a comment to explain the reason)? I want to make sure we don't accidentally lose this fix in the future.

hugobgranja:
Glad i could help, i'll create the test soon.

jkowens:
This has just been released in v0.17.2 💥

Issue #101, opened 2013-06-16
Support join table inserts


christopherchiu:
I don't mind making a second call to the library but since the import method is tied to a specific model there's no model to make the call to. 

Example: If I have store and customers, there's no way for me to insert into stores_customers even if I have both ids. 

My proposed solution would be to call ActiveRecord::Base import some_columns, some_rows, table_name: :stores_customers.

philipgiuliani:
I would need the same currently. What i did was changing `has_and_belongs_to_many` to a `has_many :abc, trough: :abc`

dcdieci:
same here. support for habtm would be appreciated.

akanshmurthy:
+1

Yes, please!

jkowens:
For future reference, another option is to create a model for the join table. So in the example above the model would be:

```
class StoresCustomer < ActiveRecord::Base
end
```

With that model you could then import the necessary records.

KyleRAnderson:
What actually is the best way of accomplishing this without support for it in this gem? I know there's [`ActiveRecord::Persistence::ClassMethods#insert_all`](https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-insert_all) but that skips all callbacks and validations.

PR #380, opened 2017-02-08
HABTM import (only for new records)

humancopy:
This pull request adds support for bulk insert of HABTM records. Its limited to new records only.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -369,6 +369,8 @@ class ActiveRecord::Base
       is_validating = options[:validate]
       is_validating = true unless options[:validate_with_context].nil?
 
+      habtm_models = []
+
       # assume array of model objects
       if args.last.is_a?( Array ) && args.last.first.is_a?(ActiveRecord::Base)
         if args.length == 2
@@ -386,6 +388,9 @@ class ActiveRecord::Base
         stored_attrs = respond_to?(:stored_attributes) ? stored_attributes : {}
 
         array_of_attributes = models.map do |model|
+          # Only do HABTM import for new records
+          habtm_models << model if model.new_record?
+
           column_names.map do |name|
             if stored_attrs.any? && stored_attrs.key?(name.to_sym)
               model.read_attribute(name.to_s)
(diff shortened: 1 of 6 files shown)


aiomaster:
Hey nice PR. It seems that some failing tests are caused by a renaming of HABTMReflection to HasAndBelongsToManyReflection in active_record the past:
https://github.com/rails/rails/commit/34c77bcf2dea31decc3a0fe78874ab598d7e8651
Maybe one should handle both cases in the `habtm_reflections` method?

humancopy:
I've fixed all issues and it now passes all tests ... :)

jkowens:
@humancopy wow thanks for taking this on. Since this is a bigger change I think I'm going to ask @zdennis to take a look when he gets a chance. This of course would only be for PostgreSQL since it supports "recursive" import. Resolves #101.

lib/activerecord-import/import.rb
```diff
@@ -633,6 +648,34 @@ def import_associations(models, options)
...
+            habtm_inserts[reflection.association_foreign_key] ||= [reflection, []]
+            habtm_inserts[reflection.association_foreign_key].last.push(*association_ids.zip([model.id] * association_ids.count))
+          end
+        end
+      end
+
+      habtm_inserts.each_value do |reflection, values|
+        join_table = reflection.options[:join_table] || reflection.join_table
+        insert_sql = prepare_insert_sql([reflection.association_foreign_key, reflection.foreign_key], table_name: join_table)
+        perform_insert insert_sql, values_sql_for_attributes(values)
```
jkowens: Just reviewing this, I'm not able to wrap my head around what `habtm_inserts` contains. Could you explain that and how many insert queries this will execute?

jkowens:
@humancopy I see that this is limited to new records only, but how should this be handled if someone combined this with the `:on_duplicate_key_update` or `:on_duplicate_key_ignore` options?

Is there any way to make this work when importing using both models and arrays of columns and values?

jkowens:
I think I would rather remove the new record only requirement and use the option `:on_duplicate_key_ignore` for HABTMs. This would mean a unique index would be necessary on the foreign key columns. This would at least prevent duplicate records.

jkowens:
Closing due to age and existing merge conflicts.

PR #381, opened 2017-02-08
Postgresql doesn't recognize column names with capital letters when returning ids.

hugobgranja:
This is caused by missing quotes around the id values when returning them.

I changed promotion_id from the promotions table to Promotion_id to trigger the error, is this enough?


hugobgranja:
It doesn't get tested in postgresql < 9.5, will test another way.

PR #382, opened 2017-02-09
Added test for column names with uppercase letters

hugobgranja:
This will test errors being caused by unquoted column names.


jkowens:
This is great. Can you just make the rubocop tweaks? Once green, I'll get these merged.

hugobgranja:
Yeah, i'm on it! 👍

hugobgranja:
Done!

PR #384, opened 2017-03-02
Add options single insert

mtparet:
#### Description ####

Add an option to force doing a single insert so we can ensure the import will be atomic without transaction or will raise an error.

### DIFF
--- a/lib/activerecord-import/adapters/mysql_adapter.rb
+++ b/lib/activerecord-import/adapters/mysql_adapter.rb
@@ -7,7 +7,7 @@ module ActiveRecord::Import::MysqlAdapter
 
   # +sql+ can be a single string or an array. If it is an array all
   # elements that are in position >= 1 will be appended to the final SQL.
-  def insert_many( sql, values, _options = {}, *args ) # :nodoc:
+  def insert_many( sql, values, options = {}, *args ) # :nodoc:
     # the number of inserts default
     number_of_inserts = 0
 
@@ -31,7 +31,7 @@ module ActiveRecord::Import::MysqlAdapter
     max = max_allowed_packet
 
     # if we can insert it all as one statement
-    if NO_MAX_PACKET == max || total_bytes <= max
+    if NO_MAX_PACKET == max || total_bytes <= max || options[:single_insert]
       number_of_inserts += 1
       sql2insert = base_sql + values.join( ',' ) + post_sql
       insert( sql2insert, *args )
--- a/test/support/mysql/import_examples.rb
+++ b/test/support/mysql/import_examples.rb
@@ -86,8 +86,11 @@ def should_support_mysql_import_functionality
     describe "single_insert" do
       let(:values) do
         # generate "big" data to go above the default mysql max_allowed_packet limit
+        max_allow_packet = Car.connection.execute("SHOW VARIABLES like 'max_allowed_packet';" ).to_a[0][1].to_i
         value = 'long_name_to_increase_size_of_the_packet_long_name_to_increase_size_of_the_packet_long_name_to_increase_size_of_the_packet_long_name_to_increase_size_of_the_packet_long_name_to_increase_size_of_the_packet_long_name_to_increase_size_of_the_packet'.freeze
-        Array.new(20_000) do
+
+        nb_values = (max_allow_packet / value.size) + 1
+        Array.new(nb_values) do
           [value]
         end
       end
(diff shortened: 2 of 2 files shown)


lib/activerecord-import/adapters/postgresql_adapter.rb
```diff
@@ -44,7 +44,7 @@ def post_sql_statements( table_name, options ) # :nodoc:
       if (options[:ignore] || options[:on_duplicate_key_ignore]) && !options[:on_duplicate_key_update] && !options[:recursive]
         sql << sql_for_on_duplicate_key_ignore( table_name, options[:on_duplicate_key_ignore] )
       end
-    elsif options[:on_duplicate_key_ignore] && !options[:on_duplicate_key_update]
+    elsif options[:on_duplicate_key_ignore] && !options[:on_duplicate_key_update] && !options[:on_duplicate_raise]
```
jkowens: I don't think this does what you are thinking. This part of the conditional just prints a warning that the installed version of PostgreSQL doesn't support the option :on_duplicate_key_ignore.

lib/activerecord-import/adapters/mysql_adapter.rb
```diff
@@ -77,7 +77,7 @@ def add_column_for_on_duplicate_key_update( column, options = {} ) # :nodoc:
       when Array then columns << column.to_sym unless columns.include?(column.to_sym)
       when Hash then columns[column.to_sym] = column.to_sym
       end
-    elsif !options[:ignore] && !options[:on_duplicate_key_ignore]
+    elsif !options[:ignore] && !options[:on_duplicate_key_ignore] && !options[:on_duplicate_raise]
```
jkowens: To disable :on_duplicate_key_update pass the option `on_duplicate_key_update: false`

jkowens:
I don't think there is a need for the :on_duplicate_raise option, if you update the PR I'll consider the option :single_insert. 

Can you add a test for the :single_insert option?

lib/activerecord-import/adapters/mysql_adapter.rb
mtparet: thanks, I'll give a try!

lib/activerecord-import/adapters/postgresql_adapter.rb
mtparet: Indeed I copied the logic although here it is not relevant.

lib/activerecord-import/adapters/mysql_adapter.rb
mtparet: I just tried and it does not raise an ActiveRecord NotUnique exception on my side.. (cf my comment below about tests)

mtparet:
Yes, I added a test for it in our project and I will put here.

lib/activerecord-import/adapters/mysql_adapter.rb
mtparet: see tests in my commit https://github.com/zdennis/activerecord-import/pull/384/commits/6c544815733a19f76bbf0b6859363f7469429519

mtparet: My bad, I tested again and it's working as attended, thanks!

mtparet:
I remove the :on_duplicate_raise option and I added a test for the single_insert option. I'm not sure if it possible to make it less "hacky".

jkowens:
I'm not sure, but I think if the import is broken into multiple inserts they are wrapped in a transaction to make sure it is atomic. From your tests is that not the case?

See: https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/adapters/mysql_adapter.rb#L43

mtparet:
Yes indeed this is the current and default behavior  for mysql adapter.
But we like to keep control if we want multiple insert inside a transaction (and auto-sizing) or not:

* We use https://github.com/basecamp/marginalia, to add automatic comment to mysql request, and this increase the size of the mysql request after active-import computed how much chunk it should import at once and so it go above the max_allowed_packet. ( perhaps we could have another fix/feature to support this case)
* Also we have to separate the chunck before using active-import because it enables us to not use a transaction around the import and so we are not locking the table during a long time.

mtparet:
> From your tests is that not the case?
As it  this is a single insert there is no transaction.

jkowens:
So after seeing what's involved, let's just forget about trying to add tests for this right now. Could you change the option name to `force_single_insert`. I think that makes the intention clear enough.

jkowens:
@mtparet I cherry picked your commit, but changed the option to `:force_single_insert`. Thanks for your work on this one. The option is now available on `master` and I'm planning on cutting a new release here shortly.

mtparet:
thanks a lot @jkowens !

jkowens:
This has just been released in v0.17.2 💥

PR #395, opened 2017-04-02
SQL Server and Oracle adapters workaround

grzegorzblaszczyk:
SQL Server and Oracle adapters workaround

### DIFF
--- a/lib/activerecord-import/base.rb
+++ b/lib/activerecord-import/base.rb
@@ -12,17 +12,22 @@ module ActiveRecord::Import
     when 'spatialite' then 'sqlite3'
     when 'postgresql_makara' then 'postgresql'
     when 'postgis' then 'postgresql'
+    when 'sqlserver' then ''
+    when 'oracle' then ''
     else adapter
     end
   end
 
   # Loads the import functionality for a specific database adapter
   def self.require_adapter(adapter)
-    require File.join(ADAPTER_PATH, "/abstract_adapter")
-    begin
-      require File.join(ADAPTER_PATH, "/#{base_adapter(adapter)}_adapter")
-    rescue LoadError
-      # fallback
+    base_adapter = base_adapter(adapter)
+    unless base_adapter.blank?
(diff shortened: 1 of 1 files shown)


jkowens:
Does an error currently occur if there is an Oracle SQLServer connection configured for the application? I thought there had been code added that prevented that from happening.

lib/activerecord-import/base.rb
```diff
@@ -12,17 +12,22 @@ def self.base_adapter(adapter)
...
-      require File.join(ADAPTER_PATH, "/#{base_adapter(adapter)}_adapter")
-    rescue LoadError
-      # fallback
+    base_adapter = base_adapter(adapter)
+    unless base_adapter.blank?
+      require File.join(ADAPTER_PATH, "/abstract_adapter")
+      begin
+        require File.join(ADAPTER_PATH, "/#{base_adapter(adapter)}_adapter")
+      rescue LoadError
+        # fallback
```
jkowens: Doesn't this rescue clause catch any errors if activerecord-import doesn't support an adapter?

grzegorzblaszczyk: Yes, you are right - it should. 

This code should also at least log some error - rescueing LoaderError without any sign does not give you any clue.

jkowens:
I still don't understand the purpose of this PR. It doesn't appear to really change any behavior. It silently skips unsupported adapters, which the existing code already does.

grzegorzblaszczyk:
I tried to merge and do a pull request on one of our old forks at https://github.com/goodylabs/activerecord-import/commit/8d2462a629dd4f5a277799186070b51ff5142b2c to have all code used by goodylabs to be potentially in your master branch.

This change was done by my collegue, Adam Migodziński, on 3 Jun 2013, so probably way before you introduced the new way of handling unsupported adapters.

I think it's safe to close this PR and forget about the whole thing. It should be already handled by a new code of yours.

jkowens:
Ohh yeah, I believe this had been fixed since then. Thanks for thinking to give back 👍

PR #406, opened 2017-04-18
Run validation callbacks

### LINKED ISSUE #405: validate callbacks does not work
`before_validation` does not work in v0.18.0.

## Test code to reproduce

```rb
# test/models/topic.rb
before_validation -> { errors.add(:title, :invalid) if title == 'invalid' }

# test/import_test.rb
it "should run callbacks" do
  assert_no_difference "Topic.count" do
    Topic.import columns, [["invalid", "Jerry Carter"]], validate: true
  end
end
```

see also: https://github.com/sinsoku/activerecord-import/commit/96fb256a28bad3fbba58754c8fe7fe867b44d7e6

jkowens:
Resolves #405.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -37,10 +37,18 @@ module ActiveRecord::Import #:nodoc:
       model.send(:validation_context=, validation_context)
       model.errors.clear
 
-      @validators.each do |v|
-        if validation_context == v.options.fetch(:on, validation_context)
-          v.validate(model) if validate?(v, model)
+      validation_proc = lambda do
+        @validators.each do |v|
+          if validation_context == v.options.fetch(:on, validation_context)
+            v.validate(model) if validate?(v, model)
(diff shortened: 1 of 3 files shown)


lib/activerecord-import/import.rb
```diff
@@ -37,12 +37,20 @@ def valid_model?(model)
...
-          v.validate(model) if validate?(v, model)
+      validation_proc = lambda do
+        @validators.each do |v|
+          if validation_context == v.options.fetch(:on, validation_context)
+            v.validate(model) if validate?(v, model)
+          end
         end
       end
 
+      if model.respond_to?(:run_validation_callbacks)
```
afn: Should this be `if model.respond_to?(:_run_validation_callbacks)`?

afn: Should this be `if model.respond_to?(:_run_validation_callbacks)`?

PR #408, opened 2017-04-19
remove numerical primary_key as well

mathieujobin:
I am hitting this bug, after switching from MySQL to PostgreSQL

` ActiveRecord::StatementInvalid: PG::NotNullViolation: ERROR: null value in column "id" violates not-null constraint DETAIL: Failing row contains (null, ...`

So I am thinking maybe we should remove the primary key column when its numerical as well.

I'm about to test this change, opening the PR early for discussion or help

thanks

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -412,7 +412,7 @@ class ActiveRecord::Base
           column_names = self.column_names.dup
         end
 
-        if models.first.id.nil? && column_names.include?(primary_key) && columns_hash[primary_key].type == :uuid
+        if models.first.id.nil? && column_names.include?(primary_key) && [:integer, :uuid].include?(columns_hash[primary_key].type)
           column_names.delete(primary_key)
         end
 


lib/activerecord-import/import.rb
```diff
@@ -412,7 +412,7 @@ def import_helper( *args )
           column_names = self.column_names.dup
         end
 
-        if models.first.id.nil? && column_names.include?(primary_key) && columns_hash[primary_key].type == :uuid
+        if models.first.id.nil? && column_names.include?(primary_key) && [:integer, :uuid].include?(columns_hash[primary_key].type)
```
zdennis: Hey @mathieujobin, thanks for catching this. I think this wasn't an issue with MySQL because MySQL treats a a `NULL` value in an INSERT statement as meaning "auto-increment please", and PostgreSQL doesn't.

Given the original condition above are you inserting some rows that _do_ have values for the `id` (primary key) column and some that do not? Or does all of your data not have values for `id`?

mathieujobin: Hey @zdennis 

thanks for the quick reply.

I've noticed a lot of this gem code assume the collection is like the first one. I guess it was the original author (you?) 's decision to make this fair assumption. conditions like this `if args.last.is_a?( Array ) && args.last.first.is_a?(ActiveRecord::Base)` I suppose it would be too much of a check to loop through the array and be counter productive in the end... 

in our case, like probably most users of this gem, we send an array of `Model.new`, so all `id` should be `nil` indeed

about having a mixed collection of ids and nils making me think of another side effect of switching to postgresql. not related to this gem, but just a FYI. setting a ID while inserting a row in postgresql will not update the sequence like mysql does this for us. you need to reset the sequence or risk to have a collision error.

jkowens:
@mathieujobin can you take a look at this comment/thread: https://github.com/zdennis/activerecord-import/issues/138#issuecomment-129631964
and see if it helps with your issue? In short some have had problems when migrating from MySQL to Postgres when their auto increment id columns did not end up getting a sequence setup properly.

mathieujobin:
@jkowens I have double check my sequence is properly owned by the table column, and `select pg_get_serial_sequence('table_name', 'id');` properly return the name of the sequence.

but inserting with `null` still makes it crash

jkowens:
Could you do one more thing? Check that in your table definition the `id` column has a default value set:

`>psql`
```
\d+ table_name
```

Under modifiers you should see something like this for the column:

```
not null default nextval('table_name_id_seq'::regclass)
```

The insert statement should be using the nextval function instead of NULL in the insert statement so that indicates something is not right at the database level.

jkowens:
Also, when checking your table definition make sure that the `id` column is defined as the primary key:

```
Indexes:
    "table_name_pkey" PRIMARY KEY, btree (id)
```

mathieujobin:
@jkowens thanks for helping trying to understand this issue

I've had checked this already, the default and primary key is properly set

```
=> \d+ subscriptions
                                                          Table "public.subscriptions"
   Column    |            Type             |                         Modifiers                          | Storage  | Stats target | Description 
-------------+-----------------------------+------------------------------------------------------------+----------+--------------+-------------
 id          | integer                     | not null default nextval('subscriptions_id_seq'::regclass) | plain    |              | 
...
Indexes:
    "subscriptions_pkey" PRIMARY KEY, btree (id)
     ....
Has OIDs: no
```

jkowens:
Hm well that's good. I'm not sure why it's not resolving the sequence value then. Here is the logic that tries to grab the sequence for the primary key:

https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/import.rb#L713

Are you able to try stepping through the code with a debugger?

mathieujobin:
thanks @jkowens  so it works perfectly.
the problem is because Amazon DMS is not recreating any indexes, or setting up those serials properly.
I was indeed missing the recommendation from #138 on my remote databases.
```
irb(main):005:0> Topic.sequence_name
=> nil
irb(main):006:0> ActiveRecord::Base.connection.execute('alter sequence topics_id_seq owned by topics.id;')
=> #<PG::Result:0x007fb73be1ddc8 status=PGRES_COMMAND_OK ntuples=0 nfields=0 cmd_tuples=0>
irb(main):007:0> Topic.sequence_name
=> "public.topics_id_seq"
irb(main):008:0> 
```

jkowens:
@mathieujobin that's great, I'm glad to hear things got worked out 👍

PR #410, opened 2017-05-03
Respect "validate" callbacks

### LINKED ISSUE #409: Validate callbacks are not called
Similar to #405, but `validate` callbacks (as opposed to `before_validation` callbacks) are still not being triggered. I'll submit a PR shortly.

# Test code to reproduce

```ruby
# test/models/topic.rb
validate -> { errors.add(:title, :invalid) if title == 'invalid' }

# test/import_test.rb
it "should run callbacks" do
  assert_no_difference "Topic.count" do
    Topic.import columns, [["invalid", "Jerry Carter"]], validate: true
  end
end
```

afn:
Currently, `before_validation` callbacks are invoked, but `validate` ones are not. This fixes #409.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -50,7 +50,16 @@ module ActiveRecord::Import #:nodoc:
               runner.call(env)
             else # ActiveRecord 5.1
               # Note that this is a gross simplification of ActiveSupport::Callbacks#run_callbacks.
-              # In particular, we don't invoke any "around_validate" callbacks.
+              # It's technically possible for there to exist an "around" callback in the
+              # :validate chain, but this would be an aberration, since Rails doesn't define
+              # "around_validate". Still, rather than silently ignoring such callbacks, we
+              # explicitly raise a RuntimeError, since activerecord-import was asked to perform
+              # validations and it's unable to do so.
+              #
+              # The alternative here would be to copy-and-paste the bulk of the
+              # ActiveSupport::Callbacks#run_callbacks method, which is undesirable if there's
+              # no real-world use case for it.
+              raise "The :validate callback chain contains an 'around' callback, which is unsupported" unless runner.final?
               runner.invoke_before(env)
               runner.invoke_after(env)
             end
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -33,28 +33,32 @@ module ActiveRecord::Import #:nodoc:
       validation_context ||= (model.new_record? ? :create : :update)
 
       current_context = model.send(:validation_context)
-      model.send(:validation_context=, validation_context)
-      model.errors.clear
+      validate_callbacks_orig = model._validate_callbacks
 
       validate_callbacks = model._validate_callbacks.dup
       validate_callbacks.each do |callback|
         validate_callbacks.delete(callback) if callback.raw_filter.is_a? ActiveRecord::Validations::UniquenessValidator
       end
 
-      model.run_callbacks(:validation) do
-        if defined?(ActiveSupport::Callbacks::Filters::Environment) # ActiveRecord >= 4.1
-          runner = validate_callbacks.compile
-          e = ActiveSupport::Callbacks::Filters::Environment.new(model, false, nil)
-          runner.call(e)
-        elsif validate_callbacks.method(:compile).arity == 0 # ActiveRecord = 4.0
-          model.instance_eval validate_callbacks.compile
-        else # ActiveRecord 3.x
-          model.instance_eval validate_callbacks.compile(nil, model)
-        end
+      begin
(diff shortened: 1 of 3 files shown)


lib/activerecord-import/import.rb
```diff
@@ -37,18 +36,21 @@ def valid_model?(model)
...
+      validate_callbacks.each do |callback|
+        validate_callbacks.delete(callback) if callback.raw_filter.is_a? ActiveRecord::Validations::UniquenessValidator
       end
 
-      if model.respond_to?(:run_validation_callbacks)
-        model.send(:_run_validation_callbacks, &validation_proc)
-      else
-        model.send(:run_callbacks, :validation, &validation_proc)
+      model.run_callbacks(:validation) do
+        if defined?(ActiveSupport::Callbacks::Filters::Environment) # ActiveRecord >= 4.1
```
afn: Is this safe? Or is there any possibility that `ActiveSupport::Callbacks::Filters::Environment` is not defined because it has not yet been auto-loaded?

lib/activerecord-import/import.rb
```diff
@@ -37,18 +36,21 @@ def valid_model?(model)
       model.send(:validation_context=, validation_context)
       model.errors.clear
 
-      validation_proc = lambda do
-        @validators.each do |v|
-          if validation_context == v.options.fetch(:on, validation_context)
-            v.validate(model) if validate?(v, model)
```
jkowens: I think you can delete the `#validate?` method below now.

lib/activerecord-import/import.rb
jkowens: I'm not entirely certain, but I think that would be safe to assume.

jkowens:
@afn thanks for working on a fix for this! I'm impressed that you figured out some of the gory validation requirements for the different versions of AR. Could you take a look at the error for AR 5.1? I just committed a change so that it is no longer an allowed failure.

lib/activerecord-import/import.rb
afn: Done

lib/activerecord-import/import.rb
```diff
@@ -24,48 +24,40 @@ def initialize(name, index)
...
+    private
 
-      Array(validator.options[:if]).map(&evaluate).compact.all? &&
-        !Array(validator.options[:unless]).map(&evaluate).compact.any?
+    def set_validate_callbacks(model, validate_callbacks)
+      if model.respond_to?(:_validate_callbacks=) # ActiveRecord < 4.1
+        model._validate_callbacks = validate_callbacks
+        model.class.__reset_runner(:validate)
+      elsif model.class.respond_to?(:_validate_callbacks=) # ActiveRecord < 5.1
+        model.class._validate_callbacks = validate_callbacks
```
jkowens: I'm a little worried this might not be thread safe. Could that be an issue?

afn: Yeah, you're right :frowning_face:

Let me rework this so it doesn't change anything at the class level. (Note that modifying the model *instance* isn't thread-safe either, but ActiveRecord is already thread-unsafe in this regard, so it's probably fine to assume that the same instance won't be simultaneously accessed from multiple threads.)

afn:
@jkowens I've reverted back to the earlier version, and added support for ActiveRecord 5.1. See comments in the code and let me know what you think. Thanks!

BTW, the build is failing in Travis for seemingly unrelated reasons.

jkowens:
@afn can you force this to run the build again? I had to update the Travis config due to some changes on their end.

jkowens:
Looks good, thanks again for digging into this issue. If you would squash all those commits now, I'll merge this into master.

afn:
Happy to squash if you prefer, but if you haven't tried it already, I'd recommend using GitHub's "squash and merge" feature. That way the PR will still show the original granular commits (useful if someone wants to drill down into the history of this change) but your master branch will only contain a single commit.

lib/activerecord-import/import.rb
```diff
@@ -24,48 +24,56 @@ def initialize(name, index)
...
+              # It's technically possible for there to exist an "around" callback in the
+              # :validate chain, but this would be an aberration, since Rails doesn't define
+              # "around_validate". Still, rather than silently ignoring such callbacks, we
+              # explicitly raise a RuntimeError, since activerecord-import was asked to perform
+              # validations and it's unable to do so.
+              #
+              # The alternative here would be to copy-and-paste the bulk of the
+              # ActiveSupport::Callbacks#run_callbacks method, which is undesirable if there's
+              # no real-world use case for it.
+              raise "The :validate callback chain contains an 'around' callback, which is unsupported" unless runner.final?
```
jkowens: I like this 👍

jkowens:
That's awesome, I need to read up on the latest Github features!

Issue #421, opened 2017-05-15
Where condition in DO UPDATE SET?


Saidbek:
Would be good to have **WHERE** condition in `up_duplicate_key_update` , for example: 

```
Project.import(
  columns,
  values,
  validate: false,
  up_duplicate_key_update: {
    conflict_target: [ :email ],
    columns: [ :first_name, ...],
    condition: "email <> 'foo@bar.com'"
  }
)
```

So far I couldn't find smth similar in source code. [postgresql doc ](https://www.postgresql.org/docs/current/static/sql-insert.html#SQL-ON-CONFLICT)

jkowens:
There's the `index_predicate` attribute for `on_duplicate_key_update`.

https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/import.rb#L338

Saidbek:
@jkowens yes I also noticed that, but it's different, there's an option (at least in PostgreSQL) where you can condition in `DO UPDATE SET` command, [DOC](https://www.postgresql.org/docs/current/static/sql-insert.html#SQL-ON-CONFLICT)

> **condition**
> An expression that returns a value of type boolean. Only rows for which this expression returns true will be updated, although all rows will be locked when the ON CONFLICT DO UPDATE action is taken. Note that condition is evaluated last, after a conflict has been identified as a candidate to update.

jkowens:
Oh I understand now. You're right, that isn't currently supported. That would be a nice feature to include and should be fairly simple to add. Are you able to work on a PR for this?

Saidbek:
Yes, I already have something working, will push soon

PR #423, opened 2017-05-18
[#421] PostgreSQL, WHERE condition to in DO UPDATE SET command

Saidbek:
**Description:**

Add an option to set WHERE condition in conflict_action, DO UPDATE SET command in PostgreSQL
[Doc](https://www.postgresql.org/docs/current/static/sql-insert.html) from PostgreSQL site.
Up #421


jkowens:
Looks good, I think the only failure is a Rubocop offense. If you could get that passing, I'll merge this into master.

jkowens:
This feature has been released in v0.19.0.

Issue #426, opened 2017-06-07
Timestamps are not imported for schemas with SQL defined default values


Andreis13:
On some of our projects, schema files are defined in SQL format instead of vanilla `schema.rb`, and it is customary for such setups to have default values on timestamp columns that are defined in the schema itself:

```sql
`created_at` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
`updated_at` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
```

In this case, `read_attribute_before_type_cast` will return `"CURRENT_TIMESTAMP"` for an unsaved record:
```ruby
Record.new.read_attribute_before_type_cast(:created_at) #=> "CURRENT_TIMESTAMP"
```

As a result, when the code reaches [this line](https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/import.rb#L788), the `array_of_attributes` is not updated with the `created_at` timestamp, because `arr[index].nil?` returns `false` (`arr[index] == "CURRENT_TIMESTAMP"` at this point). Therefore, this will import `NULL` into this column and fail because of the `NOT NULL` constraint.

jkowens:
Ah interesting. So after looking into this issue, it seems we'd probably have to use `read_attribute` instead of `read_attribute_before_type_cast` for any column that has a default value defined? It looks like PostgreSQL allows a function or expression to be set as a default value even for columns other than dates/datetimes. Does that seem like the only safe option?

Andreis13:
I don't have enough context on the behaviour of various RDBMs, in order to assess the 'safety' of this solution.. But I was wondering what is the performance difference between using `read_attribute` vs `read_attribute_before_type_cast`? Maybe it makes sense to use the former for all columns regardless of type? (might greatly simplify the existing logic)

jkowens:
From what I've seen `read_attribute_before_type_cast` is significantly faster. Here was one benchmark:

https://gist.github.com/aquajach/f875e13773ace44d6f85

I've run benchmarks where the difference is even greater.

Andreis13:
Well yeah, the method itself is surely faster, but the question is what's the difference when used in the context of `AR::Import`? I doubt it's the bottleneck of the whole process. Say, if inserting 1000 records takes 0.5 seconds, while reading 1000 times ~25 attributes with typecasting is only 0.005 seconds faster than without typecasting, we could say that this difference is negligible.

[...]

PR #428, opened 2017-06-12
Use read_attribute for columns with default value

jkowens:
Databases can allow expressions such as `CURRENT_TIMESTAMP` or functions for default values, so `read_attribute` must be used instead of `read_attribute_before_type_cast`.


jkowens:
This resolves #426.

PR #433, opened 2017-06-29
Allow RETURNING to be specified for PostgreSQL

tjwp:
The column or columns to return are specified using a `:returning` option, e.g.

```ruby
Book.import(books, returning: %w(author_name title))
```

The `ActiveRecord::Import::Result` struct is modified to include a `results` array attribute that contains the values for the returning columns.

If the table has a primary key, then the primary key fields are still requested and returned in the `ids` attribute.

### DIFF
--- a/lib/activerecord-import/adapters/abstract_adapter.rb
+++ b/lib/activerecord-import/adapters/abstract_adapter.rb
@@ -16,7 +16,7 @@ module ActiveRecord::Import::AbstractAdapter
       sql2insert = base_sql + values.join( ',' ) + post_sql
       insert( sql2insert, *args )
 
-      [number_of_inserts, []]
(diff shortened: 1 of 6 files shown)


jkowens:
Hey, thanks @tjwp. This looks like a nifty feature. @zdennis could you take a look at this when you get a chance and post your thoughts?

chewi:
Nice work. What would be even nicer is if `set_attributes_and_mark_clean` could be enhanced to feed these results back into model objects like it currently does for the primary key.

tjwp:
@chewi I've expanded this PR to set the results from returning on the model objects. For now I've kept it as a separate commit.

jkowens:
@tjwp thanks for contributing! I'll try to cut a new release sometime soon.

tjwp:
Awesome. Thanks!

jkowens:

PR #434, opened 2017-06-29
Mark clean after assigning attributes

tjwp:
This looks like it was a regression at some point.

Ids and timestamps must be assigned to models prior to marking the models as clean.


jkowens:
Good catch, thanks!

jkowens:
This has been released in v0.19.1 🎉

Issue #437, opened 2017-07-17
max_allowed_packet fails to account for single fields exceeding packet size


saizai:
https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/adapters/mysql_adapter.rb#L33

If any single field (e.g. a longtext) is larger than `max_allowed_packet`, this fails with a non-obvious error.

I would suggest either
a) raise an error specifically saying that the field X is size Y so you should run command Z on mysql to increase max_allowed_packet, or
b) just increase max_allowed_packet on the spot, and log a WARN about having done so.

jkowens:
I would be more comfortable with something along the lines of choice "a". I've opened #439 which will now raise an `ActiveRecord::Import::ValueSetTooLargeError` when an insert for a single value set will exceed `max_allowed_packet`. Hopefully it would be a more obvious error than the current one. Once I merge that in, it should be rather simple to update the MySQL adapter to rescue that error and raise a runtime error with some db specific instructions. Any interest in working on that?

jkowens:
Just an update: v0.19.1 has been released and includes the change in #439.

jkowens:
I'm thinking the new exception will provide sufficient information to figure out this issue so I'm going to close this for now. A PR is welcome if anyone feels inspired 😄

PR #439, opened 2017-07-17
Raise error when value set byte size exceeds max allowed

jkowens:
It is possible that an insert with a single value set could exceed the max allowed byte size. For MySQL that would be the value for the `max_allowed_packet setting`. This raises a more helpful error when that is the case. See #437.


jkowens:
This has been released in v0.19.1 🎉

PR #440, opened 2017-07-20
Fixed concat with 'primary_key'.

keeguon:
Hi,

In most cases a call to `primary_key` on an ActiveRecord::Base child class (model) will return a String instance which cannot be concatenated with an Array so we have to cast it into one prior to the `concat` call.


jkowens:
Good catch! Thanks for the fix 👍

keeguon:
@jkowens any release plans on rubygems soon in order to avoid fetching from git with bundler?

jkowens:
Yup, I'm working on that now and will provide an update when ready.

keeguon:
@jkowens Cool thanks ;)

jkowens:
Version 0.19.1 has been released!

PR #444, opened 2017-09-01
Fix bypassing uniqueness validators

vmaxv:
When `Validator` is iterating over a callback chain and skips unique validations, it shifts the index whenever it deletes such item from the collection. If there are several successive uniqueness validators, it omits the one following the deleted one in the chain. That prevents from inserting data in such model when running with `:validate` option.

Do not mutate the target chain inside the enumerator and iterate over a copy instead.

Relates to https://github.com/zdennis/activerecord-import/pull/301.
Follow-up for https://github.com/zdennis/activerecord-import/pull/410.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -38,7 +38,7 @@ module ActiveRecord::Import #:nodoc:
         model.errors.clear
 
         validate_callbacks = model._validate_callbacks.dup
-        validate_callbacks.each do |callback|
+        model._validate_callbacks.each do |callback|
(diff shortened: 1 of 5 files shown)


vmaxv:
Side effect of deleting elements while iterating over a collection is shown below.
```ruby
arr = (0..5).to_a
=> [0, 1, 2, 3, 4, 5]
arr.each { |n| arr.delete(n) if n.positive? }
=> [0, 2, 4]
```

jkowens:
Excellent, thanks!

jkowens:

PR #446, opened 2017-09-06
Prevent type casting for binary columns on PG to fix #443

### LINKED ISSUE #443: Postgres bytea value integrity broken on import
Hi there,

I seem to have uncovered a bug with how values are quoted when it comes to PG's binary "bytea" column. Basically upon import, the integrity of our encrypted data is broken as a result.

I've traced the problem line to here: ActiveRecord::Import:L756: `connection_memo.quote(column.type_cast(val), column)`

The issue is that `column.type_cast(val)` is calling PGconn.unescape_bytea on PG bytea column values (http://api.rubyonrails.org/v4.1.16/classes/ActiveRecord/ConnectionAdapters/PostgreSQLAdapter/OID/Bytea.html).

If we have an encrypted string, e.g.: `enc_str = "\xE0c\xB2\xB0\xB3Bh\\\xC2M\xB1m\\I\xC4r"`

and call `PGconn.unescape_bytea(enc_str)`, we get:

`"\xE0c\xB2\xB0\xB3Bh\xC2M\xB1mI\xC4r"`

The issue here is that multiple adjacent slashes (e.g. `\\\xC2M`) are removed (becoming e.g. `\XC2M`), which breaks integrity of the encryption .

Then `connection_memo.quote(...)` calls `PGconn.escape_bytea(...)` with the the mangled encryption string, which results in a successful save, but the encrypted value is no longer valid.

To work around this, we basically need to preemptively escape the encryption string before ActiveRecord::Import unescapes/escapes it. This means each encrypted value is escaped/unescaped/escaped though, which is obviously inefficient.

I guess I'm wondering if it's necessary to call `column.type_cast(val)` in the aforementioned line? 

If that can't be removed, can you think of a better workaround?

Thanks for the help :)

mwalsher:
Let me know if I missed anything here.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -772,7 +772,9 @@ class ActiveRecord::Base
               if serialized_attributes.include?(column.name)
                 val = serialized_attributes[column.name].dump(val)
               end
-              connection_memo.quote(column.type_cast(val), column)
+              # Fixes #443 to support binary (i.e. bytea) columns on PG
+              val = column.type_cast(val) unless column.type.to_sym == :binary
+              connection_memo.quote(val, column)
             end
           end
         end
--- a/test/schema/postgresql_schema.rb
+++ b/test/schema/postgresql_schema.rb
@@ -36,6 +36,7 @@ ActiveRecord::Schema.define do
     t.column :alarm_type, :integer, null: false
     t.column :status, :integer, null: false
     t.column :metadata, :text
+    t.column :secret_key, :binary
     t.datetime :created_at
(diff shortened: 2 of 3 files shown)


jkowens:
@mwalsher I think I broke the build while you were working on this PR. Are you going to add that test back in? Otherwise looks good 👍

mwalsher:
Sorry, I meant to move the test not remove it 😆 

Done.

jkowens:
Awesome, I'll update this once a new release is cut.

jkowens:

PR #448, opened 2017-09-10
Sync belongs_to association id before importing

jkowens:


jkowens:

PR #449, opened 2017-09-10
Reset model instance on validate

vmaxv:
When loading an array of hashes, the importer uses the same model instance whenever performs validations for each item. However, if there is a validator defined for a memoized attribute, it might provide false positives for valid values followed by invalid ones.

Instantiate a new object before validating a model to avoid such gotchas.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -570,8 +570,8 @@ class ActiveRecord::Base
 
         # keep track of the instance and the position it is currently at. if this fails
         # validation we'll use the index to remove it from the array_of_attributes
-        model = new
         arr.each_with_index do |hsh, i|
+          model = new
           hsh.each_pair { |k, v| model[k] = v }
           next if validator.valid_model? model
           raise(ActiveRecord::RecordInvalid, model) if options[:raise_error]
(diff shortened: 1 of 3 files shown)


jkowens:
Ah an interesting bug. I was hoping to avoid the performance hit of initializing N number of ActiveRecord objects for validation, but I guess there might be no way around it?

vmaxv:
There's always a way - if you're committed 🙂

I tried an alternative approach with marshalling:
```ruby
n = 100_000
model = Topic.new
dump = Marshal.dump(model)

Benchmark.bmbm do |x|
  x.report('new:') do
    n.times { model = Topic.new }
  end

  x.report('Marshal.load:') do
    n.times { model = Marshal.load(dump) }
  end
end
```
Here are benchmark results:
```
Rehearsal -------------------------------------------------
new:            0.770000   0.010000   0.780000 (  0.775444)
Marshal.load:   5.590000   0.020000   5.610000 (  5.629170)
---------------------------------------- total: 6.390000sec

                    user     system      total        real
new:            0.750000   0.000000   0.750000 (  0.751049)
Marshal.load:   5.680000   0.030000   5.710000 (  5.721132)
```
`new` performs way better.

We could also try to reset instance variables explicitly, i.e. remove new and restore existing ones, but that is a bit overkill and seems hacky to me.

Hence, although initializing N model instances does cause some performance regression, it is insignificant compared to the total runtime. As for memory management, we don't keep references to those one-off instances and with GC we practically don't have to worry about that.

So I'd stick with the initial solution. How do you think?

jkowens:
Yup, I think you are right. Simplicity wins out here for sure. I'm going to merge this in as is. Thanks!

vmaxv:
Just to get an idea, do you have any plans to release recent changes?

jkowens:
Yes, I think we are at a good point for a release. I'll try to get to that tomorrow.

vmaxv:
Awesome, thanks!

jkowens:
This has been released in v0.20.0 🙌

PR #465, opened 2017-10-12
Add option to fail on unknown hash keys

mbell697:
Adds an option to raise an error when passing an array of hashes to `import` and a hash contains keys which are not in `column_names`. Also clarifies in the docs that when passing only an array of hashes to `import` without explicit column names, the first hash's keys will be used as the columns to import.

The particular scenario that motivated this was using `activerecord-import` to migrate data from mongodb to postgres. It's not uncommon in mongo land for some document to not have a give key at all rather than setting the key to `null`. The result of passing mongo docs like this to `import` is that if you happen to have a doc first in the array that is missing some keys, but later docs in the same import do have those keys, they are silently ignored.  This can lead to spending several hours trying to figure out why a seemingly random assortment of column data failed to appear in postgres :)

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -247,6 +247,9 @@ class ActiveRecord::Base
     #  BlogPost.import posts
     #
     #  # Example using array_of_hash_objects
+    #  # NOTE: column_names will be determined by using the keys of the first hash in the array. If later hashes in the
+    #  # array have different keys an exception will be raised. If you have hashes to import with different sets of keys
+    #  # we recommend grouping these into batches before importing.
     #  values = [ {author_name: 'zdennis', title: 'test post'} ], [ {author_name: 'jdoe', title: 'another test post'} ] ]
     #  BlogPost.import values
     #
@@ -470,12 +473,18 @@ class ActiveRecord::Base
         if args.length == 2
           array_of_hashes = args.last
           column_names = args.first.dup
+          allow_extra_hash_keys = true
         else
           array_of_hashes = args.first
           column_names = array_of_hashes.first.keys
+          allow_extra_hash_keys = false
         end
 
         array_of_attributes = array_of_hashes.map do |h|
+          error_message = validate_hash_import(h, column_names, allow_extra_hash_keys)
+
+          raise ArgumentError, error_message if error_message
+
           column_names.map do |key|
             h[key]
           end
@@ -845,6 +854,43 @@ class ActiveRecord::Base
     # Returns an Array of Hashes for the passed in +column_names+ and +array_of_attributes+.
     def validations_array_for_column_names_and_attributes( column_names, array_of_attributes ) # :nodoc:
       array_of_attributes.map { |values| Hash[column_names.zip(values)] }
+    end
+
+    # Checks that the imported hash has the required_keys, optionally also checks that the hash has
+    # no keys beyond those required when `allow_extra_keys` is false.
+    # returns `nil` if validation passes, or an error message if it fails
+    def validate_hash_import(hash, required_keys, allow_extra_keys) # :nodoc:
+      extra_keys = allow_extra_keys ? [] : hash.keys - required_keys
+      missing_keys = required_keys - hash.keys
+
+      return nil if extra_keys.empty? && missing_keys.empty?
+
+      if allow_extra_keys
+        <<-EOS
+Hash key mis-match.
+
+When importing an array of hashes with provided columns_names, each hash must contain keys for all column_names.
+
+Required keys: #{column_names}
+Missing keys: #{missing_keys}
+
+Hash: #{hash}
+        EOS
+      else
+        <<-EOS
+Hash key mis-match.
+
+When importing an array of hashes, all hashes must have the same keys.
+If you have records that are missing some values, we recommend you either set default values
+for the missing keys or group these records into batches by key set before importing.
+
+Required keys: #{column_names}
+Extra keys: #{extra_keys}
+Missing keys: #{missing_keys}
+
+Hash: #{hash}
+        EOS
+      end
     end
   end
 end
--- a/test/import_test.rb
+++ b/test/import_test.rb
@@ -86,6 +86,54 @@ describe "#import" do
         assert_nil t.author_email_address
       end
     end
+
+    context "with extra keys" do
+      let(:values) do
+        [
+          { title: "LDAP", author_name: "Jerry Carter" },
+          { title: "Rails Recipes", author_name: "Chad Fowler", author_email_address: "cfowler@test.com" } # author_email_address is unknown
+        ]
+      end
+
+      it "should fail when column names are not specified" do
+        err = assert_raises ArgumentError do
+          Topic.import values, validate: false
+        end
+
+        assert err.message.include? 'Extra keys: [:author_email_address]'
+      end
+
+      it "should succeed when column names are specified" do
+        assert_difference "Topic.count", +2 do
+          Topic.import columns, values, validate: false
+        end
+      end
+    end
+
+    context "with missing keys" do
+      let(:values) do
+        [
+          { title: "LDAP", author_name: "Jerry Carter" },
+          { title: "Rails Recipes" } # author_name is missing
+        ]
+      end
+
+      it "should fail when column names are not specified" do
+        err = assert_raises ArgumentError do
+          Topic.import values, validate: false
+        end
+
+        assert err.message.include? 'Missing keys: [:author_name]'
+      end
+
+      it "should fail on missing hash key from specified column names" do
+        err = assert_raises ArgumentError do
+          Topic.import %i(author_name), values, validate: false
+        end
+
+        assert err.message.include? 'Missing keys: [:author_name]'
+      end
+    end
   end
 
   unless ENV["SKIP_COMPOSITE_PK"]


jkowens:
Thanks @mbell697. Importing with arrays of hashes is a relatively new addition, so I'm open to opinions on how it should work by default. I went with what I thought would be a straight forward default, but I can see how it could cause some confusion. Should we consider iterating through all of the hashes to get all possible key values up front? Does that make more sense as a default?

cc: @zdennis

zdennis:
Thanks @mbell697 for the contribution! 

> @mbell697: Adds an option to raise an error when passing an array of hashes to import and a hash contains keys which are not in column_names. 

I am leaning towards this becoming the default. Then we don't have to add an extra option. I think there is some onus of responsibility on the caller to know what's being imported and what isn't. If they try to import a column that doesn't exist that seems like an error. This way, there can be no surprises and we don't have try to guess intent.

> @mbell697: Also clarifies in the docs that when passing only an array of hashes to import without explicit column names, the first hash's keys will be used as the columns to import.

Good catch and thanks for adding this @mbell697. 

> @jkowens: Should we consider iterating through all of the hashes to get all possible key values up front? 

The thing I don't love about this is that we then need to figure out how to handle all of the hashes that don't have those keys, and while `nil` may be a good default, it seems to that on the surface it may also just be an error where activerecord-import is being told to import incompatible sets of key/value pairs (since the underlying DB import requires all columns to be the same for each value set).

One suggestion that doesn't require a change: The caller could group their hashes by keys and then run multiple `import` calls, one for each set of keys/columns. This is slightly for work for the caller but puts the caller in the situation where it's clear what their code is intending to do and that the data is expected to not conform to a homogenous set of keys/columns.

Then the check for homogenous keys/columns becomes defaulting to the first and then raise an error if any set after differs.

@jkowens / @mbell697, WDYT?

mbell697:
I think the user expectation is that passing an array of hashes to `import` would have roughly the same impact as calling `create` in a loop with the exclusion of callback / validation logic as already covered in the docs. I don't think there is any way to implement this with a single multi-row insert statement. The options for dealing with missing keys are basically 'set to nil' and 'set to database default for column', both of which are not always what the user wants, particularly if upsert use cases are considered. 

I think that @zdennis's idea of grouping the data by key set is probably the best approach. I'd probably lean toward providing helpful error messages vs doing the grouping internally as the user may have a lot more options for handling it efficiently externally. In my case I was inserting several million records in batches of 10k, depending on the key presence each batch could be broken up into 5-10 inserts internally but externally I could do a number of things to deliver batches that all have the same keys which could speed things up a bit. Alternatively ARI could also do the grouping by default but print warnings when it needs to split batches.

For my use case the most direct fix wasn't grouping by keys but doing something like this to get the full set of keys for each model pre-loaded with database / after_initialize defaults. But this is likely not something that would work for an upsert use case and likely not useful to have in ARI other than maybe as an example in the docs.

```ruby
model_attributes = MyModel.new.attributes.to_h
MyModel.import array_of_hashes.map { |h| model_attributes.merge(h) }
```

jkowens:
All good points, I think raising a helpful error message by default is the way I'd lean too. Adding some examples of how to group hashes by key when importing to the wiki and eventually the readme would be tops.

jkowens:
@zdennis do you want to make an executive decision on which route to take? 😉

jkowens:
@mbell697 I'm thinking of getting a new release out soon. Would you be up for revising this PR to raise an error when there are differences in keys and then provide helpful info (grouping, etc)?

mbell697:
@jkowens Sure. I'll get it updated today or tomorrow.

zdennis:
Sorry for the late reply @jkowens and @mbell697. 

@jkowens, I think what you mentioned earlier today is the best place to start.

jkowens:
@zdennis great I figured that would be a safe starting point 👍🏻

mbell697:
I've updated the PR but am seeing a non-sensical error message from jruby on CI and locally.

@jkowens / @zdennis any idea what the cause may be? I can't seem to find any connection to the code changes.

jkowens:
This has to be due to some change with Travis. It even happens on master branch now, and it was fine as of last week.

mbell697:
I don't think it's travis, it's failing oddly for me locally as well, note that there isn't a copy/paste error below, `File does not exist:` is not reporting the file name:

```
❯ rake test:sqlite3 --trace       
** Invoke test:sqlite3 (first_time)
** Execute test:sqlite3
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/composite_primary_keys-8.1.6/lib/composite_primary_keys/connection_adapters/abstract_mysql_adapter.rb:18: warning: `*' interpreted as argument prefix
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/composite_primary_keys-8.1.6/lib/composite_primary_keys/validations/uniqueness.rb:14: warning: shadowing outer local variable - value

File does not exist:

rake aborted!
Command failed with status (1): [ruby -w -I"lib" -I"/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib" "/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/rake_test_loader.rb" "test/adapters/sqlite3.rb" "test/import_test.rb" "test/synchronize_test.rb" "test/value_sets_bytes_parser_test.rb" "test/value_sets_records_parser_test.rb" "test/sqlite3/import_test.rb" ]
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/testtask.rb:130:in `block in define'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/file_utils.rb:57:in `sh'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/file_utils.rb:105:in `ruby'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/testtask.rb:117:in `block in define'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/file_utils_ext.rb:59:in `verbose'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/testtask.rb:111:in `block in define'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/task.rb:251:in `block in execute'
org/jruby/RubyArray.java:1734:in `each'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/task.rb:251:in `execute'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/task.rb:195:in `block in invoke_with_call_chain'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/stdlib/monitor.rb:214:in `mon_synchronize'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/task.rb:188:in `invoke_with_call_chain'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/task.rb:181:in `invoke'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/application.rb:153:in `invoke_task'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/application.rb:109:in `block in top_level'
org/jruby/RubyArray.java:1734:in `each'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/application.rb:109:in `block in top_level'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/application.rb:118:in `run_with_threads'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/application.rb:103:in `top_level'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/application.rb:81:in `block in run'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/application.rb:179:in `standard_exception_handling'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/lib/rake/application.rb:78:in `run'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/lib/ruby/gems/shared/gems/rake-12.1.0/exe/rake:27:in `<main>'
org/jruby/RubyKernel.java:973:in `load'
/Users/mark/.rbenv/versions/jruby-9.1.14.0/bin/rake:23:in `<main>'
Tasks: TOP => test:sqlite3
```

Perhaps a JVM change? Here are my versions:

```
❯ ruby --version
jruby 9.1.14.0 (2.3.3) 2017-11-08 2176f24 Java HotSpot(TM) 64-Bit Server VM 25.102-b14 on 1.8.0_102-b14 +jit [darwin-x86_64]
```

```
 ❯ java -version
java version "1.8.0_102"
Java(TM) SE Runtime Environment (build 1.8.0_102-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.102-b14, mixed mode)
```

jkowens:
You’re right...makes me think there must have been a gem updated that isn’t cooperating. Now to figure out which one 😜

lib/activerecord-import/import.rb
```diff
@@ -846,5 +855,42 @@ def add_special_rails_stamps( column_names, array_of_attributes, options )
...
+    # no keys beyond those required when `allow_extra_keys` is false.
+    # returns `nil` if validation passes, or an error message if it fails
+    def validate_hash_import(hash, required_keys, allow_extra_keys) # :nodoc:
+      extra_keys = allow_extra_keys ? [] : hash.keys - required_keys
+      missing_keys = required_keys - hash.keys
+
+      return nil if extra_keys.empty? && missing_keys.empty?
+
+      if allow_extra_keys
+        <<-EOS
```
mbell697: Just a note: I didn't know what your official Ruby version support was, if 2.3+ then `~>>` could clean this up a bit.

jkowens: We haven’t officially dropped support for Rails 3.2, so  I think that means we support Ruby 2.2.

jkowens:
@mbell697 this looks great, I'm merging and I'll post an update once I get a new release cut. Thanks!

jkowens:
This feature has been released in v0.21.0 🎉

koenhandekyn:
it was mentioned above that it would be helpful to add to the wiki a reference to a hint on grouping hashes by keys. i'm willing to add it if you can give a rough pointer.

further i have i think a use case where this validation is actually annoying. i'm using the code for both insert as update logic and in the case of updates, one often has partial updates where you don't want provide values for columns you are not updating (because it would force you to read a lot of data first). so maybe one could consider that this test is dropped in the case of upsert or by means of an explicit parameter.

unless anyone can suggest an alternative approach for doing batch partial upserts. 
thanks for your thoughts.

PR #470, opened 2017-10-20
Load correct foreign key value when syncing associations

### LINKED ISSUE #469: custom foreign_key replaced with associated records primary key in belongs_to
I discovered after #448 that a model persisted with import that has a custom foreign_key in a belongs_to association is being replaced with the primary key of the associated record.

```ruby
class Category < ApplicationRecord
  has_many :products, foreign_key: :ref_id, primary_key: :ref_id
end

class Product < ApplicationRecord
  belongs_to :category, foreign_key: :ref_id, primary_key: :ref_id
end
```

Given a category row with:

| id | ref_id |
| -- | ----------- |
| 5bffa9e2-cd56-471e-b69b-8773da1d7b4f | ref_456 | 

Before #448, `Product.import` would save the product record with:

| id | ref_id |
| -- | ----------- |
| 5ca21e1e-36b1-47c2-a182-4bc7859b1980 | ref_456 |

Now it is being saved with

| id | ref_id |
| -- | ----------- |
| 5ca21e1e-36b1-47c2-a182-4bc7859b1980 | 5bffa9e2-cd56-471e-b69b-8773da1d7b4f |

This makes `product.category` return `nil` instead of the associated record.

The following line is the culprit:
```ruby
model.public_send("#{association_reflection.foreign_key}=", association.id)
```
I am wondering why was this introduced?

I am monkey patching with the following to disable this at the moment:

```ruby
class ActiveRecord::Base
  class << self
    def load_association_ids(_model); end
  end
end
```

jkowens:
Fixes #469.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -713,8 +713,9 @@ class ActiveRecord::Base
       association_reflections.each do |association_reflection|
         association = model.association(association_reflection.name)
         association = association.target
-        if association && association.id
-          model.public_send("#{association_reflection.foreign_key}=", association.id)
+        if association
+          association_primary_key = association_reflection.association_primary_key
+          model.public_send("#{association_reflection.foreign_key}=", association.send(association_primary_key))
         end
       end
     end


marcgreenstock:
Hi @jkowens, my tests with this branch all pass. 👍 Thanks for looking into it so quickly.

Issue #467, opened 2017-10-18
Values as SQL queries


jalada:
We <3 `activerecord-import` at Pixie Labs!

We're using it to do some pretty hefty importing (running regularly, not one-off) of a whole bunch of models, a lot of them related to each other. And we've found a situation where it's better for us to construct SQL commands ourselves rather than use `activerecord-import` because it's missing one feature.

We'd like to assign some foreign keys to our imported models without looking them up in a separate query, by using a subquery for the value of a field, e.g.:

```sql
INSERT INTO comments (content, post_id) VALUES (
  ('this is content for a comment', (SELECT id FROM posts where slug = 'my-post' LIMIT 1))
)
```

The issue (I think) stopping us from doing this with `activerecord-import` is that the subquery mustn't be quoted.

The less efficient alternative is to build a big hash table in Ruby, for looking up IDs by slug when constructing the objects to pass to `activerecord-import`.

We're happy to work on this as a feature, but:

1. is this something you've thought about before? And if so, is there a reason why it might not be a good idea?
2. what could the API look like? How would you indicate to `activerecord-import` that you are passing a SQL statement? Pass it an `ActiveRecord::Relation` for `activerecord-import` to call `.to_sql` on?

If there's some other functionality in `activerecord-import` we're missing that will help us, please do say! :)

jkowens:
Glad to hear you've found this gem useful!

So the example SQL you've posted is definitely interesting. I'm not sure it's a scenario that has been discussed before. Looking at it though, I'm wondering about the slug value. It's a unique identifier in itself that you need to have somehow available for each comment to lookup the post ID. 

How are you mapping each slug? I'm just trying to wrap my head around that to get an idea how useful this would be to others. Thanks!

jalada:
We're importing data from another data source, and the data there is already loosely relational. So in my example, the data source would be providing us with a bunch of comment data: content and the slug of the post its a comment on. The posts will already have been imported and the slug stored (as a 'remote slug', perhaps), so we need to look up those posts by slug in order to create the relation in our DB.

There is an alternative, which is perhaps where you're going; the slug is a unique identifier so why not actually use that to construct the foreign key relation and do all the lookups, instead of creating another integer primary key that requires looking up? That's simply where my example breaks down but I'm sure you can envisage situations where that isn't suitable.

jkowens:
[...]

PR #471, opened 2017-10-24
Allow SQL subqueries to be passed as values

jkowens:
Objects that respond to `.to_sql` (`ActiveRecord::Relation`) can be passed as values when importing
arrays or hashes. See #467.


jalada:
I gave this a go and it worked as expected :+1:

PR #487, opened 2018-01-10
Return inserted ids for mysql bulk insert

a2ikm:
This counts only inserted records for `ON DUPLICATE KEY UPDATE` by comparing the number of inserting records and `ROW_COUNT()`.

This works good with innodb_autoinc_lock_mode = 0 and 1, but not with 2.
ref https://dev.mysql.com/doc/refman/5.7/en/innodb-auto-increment-handling.html#innodb-auto-increment-lock-modes and #279.


jkowens:
@a2ikm thanks for your interest in contributing to this project. Unfortunately, because this requires specific settings for `innodb_autoinc_lock_mode`, doesn't work with MyISAM, and won't work with `uuid` primary keys, I don't think we'll go in this direction. 

See also zdennis/activerecord-import/pull/279.

If you want to add information about calculating IDs after import in the wiki that would be a nice addition.

PR #489, opened 2018-01-16
Fix validation logic for recursive import

### LINKED ISSUE #488: N+1 with nested validation
I am importing records recursively, and a nested record has a validation that uses one of its other associations. Something like this:
```ruby
class Parent < ApplicationRecord
  has_many: children
end

class Child < ApplicationRecord
  belongs_to: parent
  belongs_to: other

  validate :something_involving_other

  def something_involving_other
    other.blah_blah_blah
  end
end

class Other < ApplicationRecord
end
```

When I bulk insert a parent with many children, it goes to the database for each child re-selecting "other", causing an N+1 issue, even if that association is cached when I pass it to `import!`.

```ruby
other = Other.first
parents = 5.times.map do
  Parent.new(children: 5.times.map { Child.new(other: other) })
end
parents.first.children.first.other # <- does not cause a db query
Parent.import!(parents, recursive: true) # <- causes 25 db queries for "other"
```

That is simplified code, of course.

Should this be happening? Can I avoid it?

jkowens:
This should resolve #488, but results in a subtle change when validating recursive imports. Only models with valid associations will be imported. Previously parent associations were imported regardless.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -592,7 +592,6 @@ class ActiveRecord::Base
         if models
           import_with_validations( column_names, array_of_attributes, options ) do |validator, failed|
             models.each_with_index do |model, i|
-              model = model.dup if options[:recursive]
               next if validator.valid_model? model
               raise(ActiveRecord::RecordInvalid, model) if options[:raise_error]
               array_of_attributes[i] = nil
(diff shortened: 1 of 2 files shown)


lib/activerecord-import/import.rb
```diff
@@ -592,7 +592,6 @@ def import_helper( *args )
         if models
           import_with_validations( column_names, array_of_attributes, options ) do |validator, failed|
             models.each_with_index do |model, i|
-              model = model.dup if options[:recursive]
```
jkowens: This seems to be the cause of the N+1 queries when models had validations that referenced associations. The associations are not included in the duped object, so they had to be loaded.

jkowens:
@eric-simonton-sama could you test out this update and let me know for sure if it resolved your issue?

eric-simonton-sama:
Yes, this fixed it!

Issue #152, opened 2014-08-08
Conflict with elasticsearch-rails gem


artworx:
Trying to use [elasticsearch-rails](https://github.com/elasticsearch/) but `import` is already used by this gem. 

Can we get something similar to [ransak](https://github.com/activerecord-hackery/ransack#ransack-search-method) where if `import` is already used, declare some other method name?

jkowens:
This would make a nice wiki entry.

PR #498, opened 2018-02-06
Rename the import method to bulk_import and alias to import.

### LINKED ISSUE #494: #import method conflicts
`elasticsearch-rails` uses the method `import` for importing data into ES. And while I am contradicted about which library holds stronger right to the `import` keyword, would you consider renaming the method to `bulk_import` or at least create an alias? I believe in general it would be nice to not use such a generic method name for an external library with such a specific function.

jkowens:
Other gems like elasticsearch-rails add an `.import` method to ActiveRecord::Base. This will resolve that name collision by allowing users to call `.bulk_import` instead. Hopefully `bulk_import` is more specific and less likely to clash with other libraries. Open to thoughts and discussion on this change.

Resolves #494 and #152.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -79,13 +79,14 @@ module ActiveRecord::Import #:nodoc:
 end
 
 class ActiveRecord::Associations::CollectionProxy
-  def import(*args, &block)
-    @association.import(*args, &block)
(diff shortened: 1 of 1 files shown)


shlima:
+1

carsol:
+1

jkowens:

Issue #472, opened 2017-10-25
Increment lock_version on duplicate key update


simonrussell:
It would be useful to automatically be able to increment the lock_version on duplicate key update.

jkowens:
This does sound like a useful feature. If anyone would like to attempt adding this functionality, a PR would be welcomed. In the mean time, a custom string of SQL can be passed for the columns value, so that might be a way to handle this situation if necessary.

simonrussell:
Indeed, I did the custom string option.  This is what I'm using:

```ruby
update_columns.map { |c| "#{c}=VALUES(#{c})" }.join(',') + ',updated_at=VALUES(updated_at),lock_version=lock_version+1'   # manually increment the lock_version and update updated_at
```

(it's ugly but it's until there's something better)

I don't have a lot of time at the moment, but I can look at making a pull request if people are interested.

PR #500, opened 2018-02-14
 Enabled to update lock_version of model

aimerald:
Hi!
Thanks for making ActiveRecord-import!!
This is great gem!

# Purpose
I wanted to solve this problem
> It would be useful to automatically be able to increment the lock_version on duplicate key update.
#472 

# did
* [x] I tested mysql!.
* [x] I tested postgres.
* [x] Updatable divided by name spaces models. (like this `Bike::Maker`)
* [x] Upsert by Models
* [x] Upsert by Arrays
* [x] Upsert by Hashes

# Approach
* When checking if there is a lock_version column at the time of update processing, update it

Sorry for my poor English.
thank you for reading! (/･ω･)/

### DIFF
--- a/lib/activerecord-import/adapters/abstract_adapter.rb
+++ b/lib/activerecord-import/adapters/abstract_adapter.rb
@@ -45,7 +45,7 @@ module ActiveRecord::Import::AbstractAdapter
       post_sql_statements = []
 
       if supports_on_duplicate_key_update? && options[:on_duplicate_key_update]
-        post_sql_statements << sql_for_on_duplicate_key_update( table_name, options[:locking_column], options[:on_duplicate_key_update], options[:primary_key] )
+        post_sql_statements << sql_for_on_duplicate_key_update( table_name, options[:on_duplicate_key_update], options[:primary_key], options[:locking_column] )
       elsif options[:on_duplicate_key_update]
         logger.warn "Ignoring on_duplicate_key_update because it is not supported by the database."
       end
--- a/lib/activerecord-import/adapters/abstract_adapter.rb
+++ b/lib/activerecord-import/adapters/abstract_adapter.rb
@@ -45,7 +45,7 @@ module ActiveRecord::Import::AbstractAdapter
       post_sql_statements = []
 
       if supports_on_duplicate_key_update? && options[:on_duplicate_key_update]
-        post_sql_statements << sql_for_on_duplicate_key_update( table_name, options[:on_duplicate_key_update], options[:primary_key] )
+        post_sql_statements << sql_for_on_duplicate_key_update( table_name, options[:locking_column], options[:on_duplicate_key_update], options[:primary_key] )
       elsif options[:on_duplicate_key_update]
         logger.warn "Ignoring on_duplicate_key_update because it is not supported by the database."
       end
--- a/lib/activerecord-import/adapters/mysql_adapter.rb
+++ b/lib/activerecord-import/adapters/mysql_adapter.rb
@@ -131,7 +131,7 @@ module ActiveRecord::Import::MysqlAdapter
   end
 
   def model(table_name)
-    table_name.delete('`').singularize.camelize.constantize
+    ActiveRecord::Base.descendants.find { |t| "`#{t.table_name}`" == table_name }
   end
 
   def locking_column?(model)
--- a/lib/activerecord-import/adapters/mysql_adapter.rb
+++ b/lib/activerecord-import/adapters/mysql_adapter.rb
@@ -84,13 +84,13 @@ module ActiveRecord::Import::MysqlAdapter
 
   # Returns a generated ON DUPLICATE KEY UPDATE statement given the passed
   # in +args+.
-  def sql_for_on_duplicate_key_update( table_name, *args ) # :nodoc:
+  def sql_for_on_duplicate_key_update( table_name, locking_column, *args ) # :nodoc:
     sql = ' ON DUPLICATE KEY UPDATE '
     arg = args.first
     if arg.is_a?( Array )
-      sql << sql_for_on_duplicate_key_update_as_array( table_name, arg )
+      sql << sql_for_on_duplicate_key_update_as_array( table_name, locking_column, arg )
     elsif arg.is_a?( Hash )
-      sql << sql_for_on_duplicate_key_update_as_hash( table_name, arg )
+      sql << sql_for_on_duplicate_key_update_as_hash( table_name, locking_column, arg )
     elsif arg.is_a?( String )
       sql << arg
     else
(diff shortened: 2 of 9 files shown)


jkowens:
@aimerald thanks for your interest in adding this new feature! I think this will be a fantastic addition. 

There are a couple of things I would like to see with this update:

1.  Look up the lock column name using the [`locking_column`](https://apidock.com/rails/ActiveRecord/Locking/Optimistic/ClassMethods/locking_column) method instead of assuming `lock_version`.
2. Find a solution that works for importing both models and an array of attribute values.

Could you see if making those adjustments would be possible? Thanks again!

aimerald:
thank you for your comment.
I think that maybe it is possible to retrieve the lock cam using `locking_column`, although it is the result of examination.

aimerald:
I verified 1 and 2 and pushed changed commits.
> Find a solution that works for importing both models and an array of attribute values.

But I can not find solution that works for importing array of attribute values...
Does anyone have a good way?

jkowens:
Could you take a look at how the `created_at`, `updated_at` timestamps are automatically added? My guess is we could potentially do something similar for `lock_version`.

aimerald:
I think that it is difficult to update if I do not know the current `lock_version` value of the object...
Is my idea wrong...??

jkowens:
Is it possible to let the database handle the increment on duplicate key? I was thinking we could build some sql that looked like `lock_version=lock_version+1`.

aimerald:
thank you for your comment!
I see. I understood
I try that way!

aimerald:
Corresponds to all upserts of models, arrays, and hashes!

aimerald:
Hum, I will check CI.

jkowens:
Awesome! It looks like you just need to implement this for Postgres and it should pass CI.

aimerald:
Postgresql test was added!
 And I also checked CI!
(/･ω･)/!

aimerald:
I found that I can not update the lock version of the model separated by namespace so I will look for a solution

aimerald:
OK!!!!
update the locking_column of the model separated by namespace!

lib/activerecord-import/adapters/mysql_adapter.rb
```diff
@@ -113,11 +114,27 @@ def sql_for_on_duplicate_key_update_as_hash( table_name, hsh ) # :nodoc:
...
+
+  def increment_locking_column!(results, table_name)
+    model = model(table_name)
+    if locking_column?(model)
+      results << "#{table_name}.`#{model.locking_column}`=`#{model.locking_column}`+1"
+    end
+  end
+
+  def model(table_name)
+    ActiveRecord::Base.descendants.find { |t| "`#{t.table_name}`" == table_name }
```
jkowens: This complexity worries me a little bit. I'm wondering if maybe it would be better if the locking column was set in the options hash and then passed down like the primary key.

aimerald: 🤔 Should I modify this code?

aimerald: Be sure to validate that model has locking_column.
In order to verify, it is necessary to obtain a list of columns held by the model, and it is necessary to find model from the table name.
Because I think it is difficult to delete this method...
If I know the model, I think locking_column is useful because it knows locking_column even if it is not specified by option.
How about that?

jkowens: What I meant was that we could insert the lock_version value into the options hash back in the initial `import_helper` method, similar to how `primary_key` is handled.  If the value is nil, we'll know they table doesn't have a lock column. The trick would be to make sure it gets passed down to these `on_duplicate_key_update` methods.

See: https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/import.rb#L481

aimerald:
Fix to pass locking_column information from import_helper
I think that this one is certainly simple!

lib/activerecord-import/adapters/postgresql_adapter.rb
```diff
@@ -118,7 +118,7 @@ def sql_for_on_duplicate_key_ignore( table_name, *args ) # :nodoc:
 
   # Returns a generated ON CONFLICT DO UPDATE statement given the passed
   # in +args+.
-  def sql_for_on_duplicate_key_update( table_name, *args ) # :nodoc:
+  def sql_for_on_duplicate_key_update( table_name, locking_column, *args ) # :nodoc:
     arg, primary_key = args
```
jkowens: We can leave this method signature the same and just do:

`arg, primary_key, locking_column = args`

lib/activerecord-import/import.rb
```diff
@@ -475,6 +475,7 @@ def import_helper( *args )
       options.merge!( args.pop ) if args.last.is_a? Hash
       # making sure that current model's primary key is used
       options[:primary_key] = primary_key
+      options[:locking_column] = locking_column if attribute_names.include?(locking_column)
```
jkowens: Is the conditional check necessary? Can we just do:

`options[:locking_column] = locking_column`

lib/activerecord-import/adapters/abstract_adapter.rb
```diff
@@ -45,7 +45,7 @@ def post_sql_statements( table_name, options ) # :nodoc:
       post_sql_statements = []
 
       if supports_on_duplicate_key_update? && options[:on_duplicate_key_update]
-        post_sql_statements << sql_for_on_duplicate_key_update( table_name, options[:on_duplicate_key_update], options[:primary_key] )
+        post_sql_statements << sql_for_on_duplicate_key_update( table_name, options[:locking_column], options[:on_duplicate_key_update], options[:primary_key] )
```
jkowens: Let's make this:

`sql_for_on_duplicate_key_update( table_name, options[:on_duplicate_key_update], options[:primary_key], options[:locking_column])`

lib/activerecord-import/adapters/mysql_adapter.rb
```diff
@@ -84,13 +84,13 @@ def add_column_for_on_duplicate_key_update( column, options = {} ) # :nodoc:
 
   # Returns a generated ON DUPLICATE KEY UPDATE statement given the passed
   # in +args+.
-  def sql_for_on_duplicate_key_update( table_name, *args ) # :nodoc:
+  def sql_for_on_duplicate_key_update( table_name, locking_column, *args ) # :nodoc:
```
jkowens: Method signature can remain `sql_for_on_duplicate_key_update( table_name, *args )`

lib/activerecord-import/adapters/mysql_adapter.rb
```diff
@@ -84,13 +84,13 @@ def add_column_for_on_duplicate_key_update( column, options = {} ) # :nodoc:
...
     sql = ' ON DUPLICATE KEY UPDATE '
     arg = args.first
     if arg.is_a?( Array )
```
jkowens: `locking_column = args.last`

aimerald: When locking_column parameters depended in order, I felt it was difficult to change in the future so I did this.

lib/activerecord-import/import.rb
aimerald: locking_column

aimerald: lock_versionを持たないモデルでも
`Topic` model has no lock_version. 
but `Topic.locking_column` returns `"lock_version"`

jkowens: Ah ok, so then your check makes sense. I was expecting models with no lock version to return nil.

aimerald:
I modified the commented place!
I think that it became a good feeling (/･ω･)/

jkowens:
Thank you! I will give it one last look in the morning with some fresh eyes, but I think this is ready to merge 👍

jkowens:
Thanks again for all your work on this!

aimerald:
I was very fortunate to have you review.
thanks for your all review!

jkowens:

Issue #496, opened 2018-02-02
on_duplicate_key_update all fields


itay-grudev:
Is there a way to update all fields in case of duplicates? Something like:

```ruby
Book.import books, on_duplicate_key_update: :all
```

itay-grudev:
Even better would be all fields except a given list. I don't know if there is native support for this, but it can be easily accomplished as we have the database schema.

jkowens:
This is an interesting idea. There currently isn't an option for this, but it's not a hard thing to calculate before import:

```
columns = Book.column_names - ["id"]
Book.import books, on_duplicate_key_update: columns
```
Figuring out a syntax for these options might be the tricky part. We're trying to keep it similar between MySQL and Postgres for basic scenarios. That being the case adding an `:all` option might be the simplest path, but I'm open to discussion.

aadityataparia:
+1

PR #506, opened 2018-02-28
usable on_duplicate_key_update :all options

aimerald:
## Purpose
I try this iseeus (/･ω･)/
https://github.com/zdennis/activerecord-import/issues/496

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -483,8 +483,8 @@ class ActiveRecord::Base
 
       # Don't modify incoming arguments
       on_duplicate_key_update = options[:on_duplicate_key_update]
-      if on_duplicate_key_update && (on_duplicate_key_update.duplicable? || on_duplicate_key_update == :all)
-        on_duplicate_key_update = attribute_names.reject { |attr| attr == primary_key }.map(&:to_sym) if on_duplicate_key_update == :all
+      on_duplicate_key_update = attribute_names.reject { |attr| attr == primary_key }.map(&:to_sym) if on_duplicate_key_update == :all
+      if on_duplicate_key_update && on_duplicate_key_update.duplicable?
         options[:on_duplicate_key_update] = if on_duplicate_key_update.is_a?(Hash)
           on_duplicate_key_update.each { |k, v| on_duplicate_key_update[k] = v.dup if v.duplicable? }
         else
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -483,7 +483,8 @@ class ActiveRecord::Base
 
       # Don't modify incoming arguments
       on_duplicate_key_update = options[:on_duplicate_key_update]
-      if on_duplicate_key_update && on_duplicate_key_update.duplicable?
+      if on_duplicate_key_update && (on_duplicate_key_update.duplicable? || on_duplicate_key_update == :all)
+        on_duplicate_key_update = attribute_names.map(&:to_sym) if on_duplicate_key_update == :all
         options[:on_duplicate_key_update] = if on_duplicate_key_update.is_a?(Hash)
           on_duplicate_key_update.each { |k, v| on_duplicate_key_update[k] = v.dup if v.duplicable? }
         else
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -484,7 +484,7 @@ class ActiveRecord::Base
       # Don't modify incoming arguments
       on_duplicate_key_update = options[:on_duplicate_key_update]
       if on_duplicate_key_update && (on_duplicate_key_update.duplicable? || on_duplicate_key_update == :all)
-        on_duplicate_key_update = attribute_names.map(&:to_sym) if on_duplicate_key_update == :all
+        on_duplicate_key_update = attribute_names.reject { |attr| attr == primary_key }.map(&:to_sym) if on_duplicate_key_update == :all
         options[:on_duplicate_key_update] = if on_duplicate_key_update.is_a?(Hash)
           on_duplicate_key_update.each { |k, v| on_duplicate_key_update[k] = v.dup if v.duplicable? }
         else
--- a/test/support/shared_examples/on_duplicate_key_update.rb
+++ b/test/support/shared_examples/on_duplicate_key_update.rb
@@ -220,7 +220,11 @@ def should_support_basic_on_duplicate_key_update
             book.tag_id = 2
             book.publisher_id = 3
             book.for_sale = false
-            book.status = 4
+            if ENV['AR_VERSION'].to_f >= 4.1
+              book.draft!
+            else
+              book.status = 0
+            end
             book
           end
           Book.import(updated_books, on_duplicate_key_update: :all)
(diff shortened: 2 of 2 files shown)


aimerald:
Ops...

lib/activerecord-import/import.rb
```diff
@@ -483,7 +483,8 @@ def import_helper( *args )
 
       # Don't modify incoming arguments
       on_duplicate_key_update = options[:on_duplicate_key_update]
-      if on_duplicate_key_update && on_duplicate_key_update.duplicable?
+      if on_duplicate_key_update && (on_duplicate_key_update.duplicable? || on_duplicate_key_update == :all)
+        on_duplicate_key_update = attribute_names.map(&:to_sym) if on_duplicate_key_update == :all
```
jkowens: You would need to exclude the primary key attribute:
```
if on_duplicate_key_update == :all 
  on_duplicate_key_update = attribute_names.map { |attr| attr.to_sym unless attr == primary_key}.compact
end
```

lib/activerecord-import/import.rb
```diff
@@ -483,7 +483,8 @@ def import_helper( *args )
 
       # Don't modify incoming arguments
       on_duplicate_key_update = options[:on_duplicate_key_update]
-      if on_duplicate_key_update && on_duplicate_key_update.duplicable?
+      if on_duplicate_key_update && (on_duplicate_key_update.duplicable? || on_duplicate_key_update == :all)
```
jkowens: I don't think we need to change this conditional, because I think it might make more sense to move the code below outside of this if block.

lib/activerecord-import/import.rb
jkowens: I think we can move this code outside of this if block down to line 494.

lib/activerecord-import/import.rb
aimerald: I think that this condition is necessary because on_duplicate_key_update argument expects an Array or Hash.

aimerald:
```
Book.import(books, on_duplicate_key_update: [:all])
```
In this way I think that it is possible to update all the columns without changing the condition, which do you think is appropriate?

aimerald:
Ops...
Maybe I think I understand why CI fails, so I fix it

test/support/shared_examples/on_duplicate_key_update.rb
```diff
@@ -213,6 +213,27 @@ def should_support_basic_on_duplicate_key_update
...
 
     context "with :on_duplicate_key_update" do
+      describe 'usable :all when on_duplicate_key_update' do
+        it 'usable :all options' do
+          updated_books = Book.all.map do |book|
+            book.topic_id = 1
+            book.tag_id = 2
+            book.publisher_id = 3
+            book.for_sale = false
+            book.status = 4
```
jkowens: I believe valid statuses are `:draft` or `:published`.

aimerald:
OK!
CI street passed!

aimerald:
>  think that this condition is necessary because on_duplicate_key_update argument expects an Array or Hash.
In this way I think that it is possible to update all the columns without changing the condition, which do you think is appropriate?

lib/activerecord-import/import.rb
jkowens: I like the way you have it with `on_duplicate_key_update: :all`. That avoids the possible issue of a table having a field named `all`.

jkowens: `if on_duplicate_key_update && on_duplicate_key_update.duplicable?` results in `false` if on_duplicate_key_update is `:all` correct?

If that's the case, I say we just move

`on_duplicate_key_update = attribute_names.reject { |attr| attr == primary_key }.map(&:to_sym) if on_duplicate_key_update == :all` directly below that if block.

aimerald: `if on_duplicate_key_update && on_duplicate_key_update.duplicable?`
This results will be false, because `on_duplicate_key_update.duplicable? ` returns an false if it is not an array or hash
`on_duplicate_key_update` has a `:all`. It is a Symbol.

aimerald:
I tried to fix so that all columns can be updated without changing conditions!

lib/activerecord-import/import.rb
```diff
@@ -483,6 +483,7 @@ def import_helper( *args )
 
       # Don't modify incoming arguments
       on_duplicate_key_update = options[:on_duplicate_key_update]
+      on_duplicate_key_update = attribute_names.reject { |attr| attr == primary_key }.map(&:to_sym) if on_duplicate_key_update == :all
```
jkowens: Can you move this down to line 494? I don't think there is a need to dup this array, which is what will happen in the next if block.

aimerald: > #import with :on_duplicate_key_update usable :all when on_duplicate_key_update#test_#import with :on_duplicate_key_update usable :all when on_duplicate_key_update_usable :all options:
ActiveRecord::RecordNotUnique: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "books_pkey"
DETAIL:  Key (id)=(1) already exists.

aimerald: #import with :on_duplicate_key_update usable :all when on_duplicate_key_update#test_#import with :on_duplicate_key_update usable :all when on_duplicate_key_update_usable :all options:
ActiveRecord::RecordNotUnique: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "books_pkey"
It happen(´・ω・｀)

jkowens: Ah my mistake, I misunderstood the code there 😜

jkowens:
Just had a thought...

What happens if someone is importing an array of arrays and doesn't supply all of the attributes?

```
columns = [ :title ]
values = [ ['Book1'], ['Book2'] ]

Book.import columns, values, on_duplicate_key_update: :all
```

Does this scenario cause any problems?

aimerald:
Results I tried Results like this are displayed

> ActiveRecord::StatementInvalid: PG::NotNullViolation: ERROR:  null value in column "author_name" violates not-null constraint
> DETAIL:  Failing row contains (67, Book1, Default Publisher, null, 2018-03-11 05:57:29.029609, 2018-03-11 05:57:29.029609, 2018-03-11 05:57:29.029609, 2018-03-11 05:57:29.029609, null, null, null, null, t, 0, null).
> : INSERT INTO "books" ("title","created_at","created_on","updated_at","updated_on") VALUES ('Book1','2018-03-11 05:57:29.029609','2018-03-11 05:57:29.029609','2018-03-11 05:57:29.029609','2018-03-11 05:57:29.029609'),('Book2','2018-03-11 05:57:29.029609','2018-03-11 05:57:29.029609','2018-03-11 05:57:29.029609','2018-03-11 05:57:29.029609')  RETURNING "id"

jkowens:
I'm thinking that might be a problem. I think a user would expect that "all" would apply to all the columns specified either using arrays or hashes.

aimerald:
Please tell me if you have a good solution for specifying columns with Hash or Array(´・ω・｀)

aimerald:
I will search for other methods to solve this problem and close!

jkowens:
Sorry, I’ve been so busy lately I haven’t had time to look into this more. Thanks for your help making activerecord-import better!

jkowens:
@octavpo yeah, we're in agreement on how it should work. I think the problem was the implementation of that is more challenging than initially thought. We need a solution that will work for importing models, arrays, and hashes.

octavpo:
But the code already deals with the complexity of determining the list of column names for the different combinations of arguments, why not just reuse it (after removing the primary key of course)? Am I missing something here?

jkowens:
No, I think that's about it. I'm totally open to someone picking this one back up and bringing it across the finish line 😉

PR #508, opened 2018-03-15
Fix import_without_validations_or_callbacks fails if array is empty

doloopwhile:
In the latest version, `#import_without_validations_or_callbacks` raises "invalid slice size" exception because #each_slice is called with 0.

`#import_without_validations_or_callbacks` is a public API (Is my understanding correct?).
So it have to work even if `array_of_attributes` is empty.


jkowens:
Good catch, thanks 👍

jkowens:

PR #512, opened 2018-05-04
Use the association primary key when importing

dpogue:
When importing into an association, it was using the owner class table primary key (usually `id`) rather than the `:primary_key` option specified on the association.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -97,7 +97,7 @@ class ActiveRecord::Associations::CollectionAssociation
     symbolized_foreign_key = reflection.foreign_key.to_sym
     symbolized_column_names = model_klass.column_names.map(&:to_sym)
 
-    owner_primary_key = owner.class.primary_key
+    owner_primary_key = reflection.active_record_primary_key.to_sym
     owner_primary_key_value = owner.send(owner_primary_key)
 
     # assume array of model objects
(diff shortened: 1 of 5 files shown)


jkowens:
Thanks for this! Any chance you could add test coverage for this case?

dpogue:
Ah, will do!

jkowens:
It's strange that the test is failing for Rails 3.2, but can you move this test to the `test/import_test.rb` file? The file it is currently in is specific testing `on_duplicate_key_update`.

test/support/shared_examples/on_duplicate_key_update.rb
```diff
@@ -284,6 +284,27 @@ def should_support_basic_on_duplicate_key_update
...
+        let(:update_columns) { [:token] }
+
+        macro(:perform_import) do |*opts|
+          @user.user_tokens.import columns, values, opts.extract_options!.merge(on_duplicate_key_update: update_columns, validate: false)
+        end
+        macro(:created_token) { UserToken.find(5) }
+
+        setup do
+          User.import [User.new(id: 1, name: 'Solomon')], validate: false
+          @user = User.find 1
```
jkowens: This could be simplified a bit:

` @user = User.create(id: 1, name: 'Solomon')`

dpogue:
Thanks! I've moved the test and it's all passing now :)

jkowens:
Looks good, thanks again 👍

jkowens:
Released in v0.24.0 👍

Issue #526, opened 2018-06-21
Support for makara_postgis adapter


chadwilken:
I see you have support for postgres makara and postgis, but not makara_postgis. Would that be as simple as adding to [base.rb](https://github.com/zdennis/activerecord-import/blob/d80417ff97d7b536705a8f6a2adfcda40666b59c/lib/activerecord-import/base.rb#L13)?

jkowens:
PR welcome btw 😃

chadwilken:
I should be able to whip one up. I hit some issues with Makara so I took a
break on this endeavor for now. Still happy to contribute though.
-- 
*Chad Wilken*
1125 Aberdeen Avenue
Lincoln, NE 68512
(402) 405-7300

PR #527, opened 2018-06-25
Add Makara Postgis adapter

chadwilken:
As discussed in #526 there isn't support for the Makara Postgis adapter, so I figured I should change that.

This branch does fail the same tests as the existing postgis and postgresql adapters with the error `duplicate key value violates unique constraint "uk_code" DETAIL:  Key (code)=(abc) already exists.`


jkowens:

PR #530, opened 2018-07-03
Ignore virtual columns

### LINKED ISSUE #521: import should not try to import virtual/generated columns
When trying to import into a model which contains a virtual column, the import try to import using all columns, including the virtual ones (generated), generating an error:

The column are created like
```ruby
t.virtual :payment_id, 
  as: "source_data->>'$.metadata.payment_id'", 
  type: :integer, 
  null: true, 
  stored: true
```
They are managed by the database (mysql in my case)

The error looks like:
`activerecord Mysql2::Error: The value specified for generated column in table`

The importer, when trying to import, should try to ignore the virtual columns.
Something like:
```ruby
MyModel.columns.reject(&:virtual?).map(&:name)
```
if the adapter support the virtual columns (there is a `supports_virtual_columns?`method)

As a workaround, I passed the list of column to update in my import.

WDYT?

jkowens:
Fixes #521.


lib/activerecord-import/import.rb
```diff
@@ -101,7 +101,12 @@ def bulk_import(*args, &block)
 
     model_klass = reflection.klass
     symbolized_foreign_key = reflection.foreign_key.to_sym
-    symbolized_column_names = model_klass.column_names.map(&:to_sym)
+
+    symbolized_column_names = if model_klass.connection.respond_to?(:supports_virtual_columns?) && model_klass.connection.supports_virtual_columns?
+      model_class.columns.reject(:virtual?).map { |c| c.name.to_sym }
```
dbourguignon: @jkowens I think you mean `model_class.columns.reject(&:virtual?).map { |c| c.name.to_sym }` here

lib/activerecord-import/import.rb
```diff
@@ -510,7 +515,11 @@ def import_helper( *args )
           column_names = args.first.dup
         else
           models = args.first
-          column_names = self.column_names.dup
+          column_names = if connection.respond_to?(:supports_virtual_columns?) && connection.supports_virtual_columns?
+            columns.reject(:virtual?).map(&:name)
```
dbourguignon: @jkowens I think you mean `columns.reject(&:virtual?).map(&:name)` here

lib/activerecord-import/import.rb
jkowens: I sure do, thanks!

jkowens:
@dbourguignon I've corrected those mistakes, hopefully this will work for you now 😃

dbourguignon:
👍 @jkowens yes it works now… good job

jkowens:

PR #531, opened 2018-07-03
Validate models before building arrays of values

### LINKED ISSUE #510: Attributes set in before_validation ignored
Morning Folks,

I am using activerecord-import to import a bunch of AR models. One of my models uses the `before_validation` hook to set an attribute on the model.

```ruby
before_validation :persist_timeline_state

private

    def persist_timeline_state
        write_attribute(:timeline_state, determine_timeline_state)
    end
```

Ordinarily when calling `save!` on the model this works as expected, however active-record-import doesn't appear to spot the new attribute value, and raises a NullConstraint error when trying to run the insert.

If I invoke the `persist_timeline_state` method manually while building each instance before running the `import!` method everything works fine.

Debugging shows that the `persist_timeline_state` is being run by `import!`, but it just seems that the newly updated attribute isn't used when running the insert.

jkowens:
Fixes #510.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -500,8 +500,8 @@ class ActiveRecord::Base
         end
       end
 
-      is_validating = options[:validate]
-      is_validating = true unless options[:validate_with_context].nil?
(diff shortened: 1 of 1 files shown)


jkowens:

PR #543, opened 2018-08-27
Add option to update all fields with on_duplicate_key_update

### LINKED ISSUE #496: on_duplicate_key_update all fields

jkowens:
Resolves #496.


aimerald:
Thank you for completing! 👍 🙏 👏

Issue #548, opened 2018-09-20
MySql :on_duplicate_key_update enabled by default with timestamp columns


jimthrive:
### Background

In the `MysqlAdapter` when `:timestamps` is set to `true` (which is default), and `:ignore` is set to `false` (also default) `options[:on_duplicate_key_update]` is set to `[updated_at]` (or whatever the updated_at timestamp column is called in your environment). This doesn't seem to be the case for Postgres.

### Effects
The main issue is that this allows duplicates in your import, even when you don't have options set for ignoring or updating on duplicates. 

### Steps to Reproduce
Using mysql, import something with duplicate keys (either via primary key, or unique index). You'll notice the `created_at` and `updated_at` have different timestamps for rows with the duplicate keys, and that the import succeeds. 

### Relevant Code

In `Import.add_special_rails_stamps()`:

```
if supports_on_duplicate_key_update? && action == :update
  connection.add_column_for_on_duplicate_key_update(column, options)
end
```

In `MysqlAdapter.add_column_for_on_duplicate_key_update()`:

```
def add_column_for_on_duplicate_key_update( column, options = {} ) # :nodoc:
  if options.include?(:on_duplicate_key_update)
    columns = options[:on_duplicate_key_update]
    case columns
    when Array then columns << column.to_sym unless columns.include?(column.to_sym)
    when Hash then columns[column.to_sym] = column.to_sym
    end
  elsif !options[:ignore] && !options[:on_duplicate_key_ignore]
    options[:on_duplicate_key_update] = [column.to_sym]
  end
end
```

### Suggested Fix
I'll submit a PR for this in a bit...

Change `MysqlAdapter.add_column_for_on_duplicate_key_update()` to only add columns if `options.include?(:on_duplicate_key_update) && !options[:ignore] && !options[:on_duplicate_key_ignore]`. 

It might be also be worth adding this check to `PostgreSQLAdapter` and cleaning up some of the mutex options, similar to what's been done in `PostgreSQLAdapter.post_sql_statements()`.

jkowens:
@jimthrive I agree this is bizarre default behavior, but I think why we never removed it was because this could break applications that now depend on it to work this way. When `on_duplicate_key_update` support was added for Postgres and SQLite, this default behavior was not included.

To work around this there is the option to do `on_duplicate_key_update: false`. See: https://github.com/zdennis/activerecord-import/pull/137

[...]

PR #549, opened 2018-09-20
Fix mysql duplicate key bug 

jimthrive:
See issue https://github.com/zdennis/activerecord-import/issues/548.

### DIFF
--- a/lib/activerecord-import/adapters/mysql_adapter.rb
+++ b/lib/activerecord-import/adapters/mysql_adapter.rb
@@ -71,7 +71,7 @@ module ActiveRecord::Import::MysqlAdapter
 
   # Add a column to be updated on duplicate key update
   def add_column_for_on_duplicate_key_update( column, options = {} ) # :nodoc:
-    if columns = options[:on_duplicate_key_update]
+    if (columns = options[:on_duplicate_key_update])
       case columns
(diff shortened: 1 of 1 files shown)


jimthrive:
Looks like a travis issue:

```
ERROR:  While executing gem ... (Gem::RemoteFetcher::UnknownHostError)
    timed out (https://api.rubygems.org/specs.4.8.gz)
```

oniofchaos:
Ideally there'd be some sort of regression test that prevents the described bug from happening again in the future

lib/activerecord-import/adapters/mysql_adapter.rb
```diff
@@ -71,14 +71,12 @@ def pre_sql_statements( options)
 
   # Add a column to be updated on duplicate key update
   def add_column_for_on_duplicate_key_update( column, options = {} ) # :nodoc:
-    if options.include?(:on_duplicate_key_update)
+    if options.include?(:on_duplicate_key_update) && !options[:ignore] && !options[:on_duplicate_key_ignore]
```
jkowens: With this change, I think we could simplify this to:

`if options[:on_duplicate_key_update] && !options[:ignore] && !options[:on_duplicate_key_ignore]`

jkowens:
This has been released in v1.0.0 🎉

PR #550, opened 2018-09-25
Allow .import to work with ProxySQL

GregFarrell:
Remove extraneous ';' from command to query mysql max_allowed_packet

There was an un-necessary semicolon at the end of the command to query
the MySQL servers max_allowed_packet global variable.
 If ProxySQL was in between activerecord-import and the database then
 this semicolon would cause "Mysql2::Error: Commands out of sync" errors.

These errors occur if multiple queries are performed without reading the
responses of all of them. The semicolon through ProxySQL caused an extra
query and these errors.

Remove the semicolon here as it is unnecessary, triggers a problem in
this particular case, and may cause issues in other cases also.


jkowens:
Thank you!

GregFarrell:
Thank you Jordan!

PR #552, opened 2018-09-26
Start the README refactor and add a section for callbacks

### LINKED ISSUE #442: before_save, before_create, before_update
My reading of the code is that importing with validation will run the `before_validation` and `validate` callbacks, but not the `before_save` and `before_create` (or `before_update`) callbacks before saving.

On the other hand, the approach in https://github.com/zdennis/activerecord-import/wiki/Callbacks will run the `before_save` and `before_create` callbacks before the validations, which isn't as expected either (http://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html)

Is there a reason why this is so, or am I misreading the code?

oniofchaos:
Closes #442

Reorganize the README to explicitly have a section dedicated to usage of
activerecord-import, similar to how the wiki is currently laid out.

Add a note that the wiki is not to be added to going forward.

Migrate the callbacks wiki article to the README, plus add the example
from #442 to add an example of how callbacks and validations can
interact in a different fashion.

@zdennis @jkowens if you like this style I can also work on the others as I have time.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -21,7 +21,7 @@ and then the reviews:
 That would be about 4M SQL insert statements vs 3, which results in vastly improved performance. In our case, it converted
 an 18 hour batch process to <2 hrs.
 
-## Usage Table of Contents
+## Table of Contents
 
 * [Callbacks](#callbacks)
 * [Additional Adapters](#additional-adapters)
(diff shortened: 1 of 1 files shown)


jkowens:
Thanks for getting something started! I just got back from vacation, so I'll try to look this over within the next couple of days.

jkowens:
This looks great 💯  It reminds me of the Sinatra readme which I was hoping to model this projects readme after.

oniofchaos:
Perfect. I should be able to start moving through the rest within the next week or so - I've been on vacation but am returning now.

Issue #507, opened 2018-03-09
Importing an array of hashes ignores columns not in the first hash


ahmad-sanad:
Title says it all. Say we have table `Foo` with columns `id`, `bar`, `baz`, and want to import the following:
```ruby
arr = [
  { bar: 'abc' },
  { baz: 'xyz' },
  { bar: '123', baz: '456' }
]

Foo.import arr
```
This creates the rows:
```
+----+------+------+
| id | bar  | baz  |
+----+------+------+
| 1  | abc  | null |
+----+------+------+
| 2  | null | null |
+----+------+------+
| 3  | 123  | null |
+----+------+------+
```

instead of the expected:
```
+----+------+------+
| id | bar  | baz  |
+----+------+------+
| 1  | abc  | null |
+----+------+------+
| 2  | null | xyz  |
+----+------+------+
| 3  | 123  | 456  |
+----+------+------+
```

It's possible to get around this by using `ActiveRecord` objects, for example:
```
arr.map! { |args| Foo.new(args) }
Foo.import arr
```
but this is still a rather pernicious bug.

yinghaochan:
points for use of `pernicious`

jkowens:
Are you using a version older than 0.21.0? I believe this issue was addressed in that release.

ahmad-sanad:
Thanks for the response, looks like I'm off by a version. Can this be documented somewhere? Since this use case isn't supported (it raises an exception), I feel like it's a worthy addition to the docs. Maybe [here](https://github.com/zdennis/activerecord-import/wiki/Examples#import-using-hashes)? I realize that isn't what the example is doing, but it doesn't explicitly say that it isn't possible either.

Also, looking through the discussion of how the decision of raising an exception came to be, I'm not sure I understand the concerns with setting the values to column defaults. That is, after all, the default behavior for ActiveRecord (or any insert query where you don't specify the value of a column for that matter) and I think a valid expectation for this gem is that it would do the same. It would nice to have that as an option at least.

I'm happy to create PRs for any of my suggestions if we are in agreement.

jkowens:
Absolutely, this should definitely be documented at that wiki page. For now, an example of how to group group hashes by keys  as suggested would be nice as well:

[...]

Issue #430, opened 2017-06-19
Document lack of uniqueness validation


xruins:
Even `validate: true` was set for `import` method,
ActiveRecord::Validations::UniquenessValidator will be disabled by the code on https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/import.rb#L42 .

Since it's confusable without an annotation, I prefer this incompatibility is mentioned in some document.

jkowens:
Good idea. The history of why uniqueness validation is avoided is found here: https://github.com/zdennis/activerecord-import/issues/228.

xRuins:
Thanks! :)

jkowens:
An option `validate_uniqueness` has been added in v0.27.0 that allows users to enable uniqueness validation:

```
Book.import books, validate_uniqueness: true
```
This is not a recommended approach as it has several potential issues, so user beware ☠️

oniofchaos:
It seems like instead of adding documentation around an option that doesn't do anything, the better approach would be to get rid of the option?

jkowens:
@oniofchaos we had completely removed uniqueness validation, but with v0.27.0 we added an option to enable it. By default validating an import will not validate uniqueness. The option `validate_uniqueness` must be set to true for it to happen. So that's really what needs documented now, the default behavior and the special option.

xruins:
Thanks a lot!

khalilgharbaoui:
I know this issue is closed but there is an unanswered question here:

does `validate_uniqueness: true` work independently from `validate: true`?

If `yes` maybe we need it in the docs if `no` we also need it in the docs
This is otherwise confusing and a pain even on and beyond 04-12-219

jkowens:
`validate_uniqueness` takes affect only when `validate: true`. It does not work independently. A PR to document this is welcome. Thanks!

khalilgharbaoui:
@jkowens are you 100% sure that it only takes effect when `validate: true` because I tried it in addition to `validate: false` and it does trigger the uniqueness validations on my model.
Does that mean it also activates and sets `validate: true` even if you explicitly set `validate: false`.

I have finally given up and used a database constraint instead. The above is very unreliable.
At least the database constraint will always raise an error.
https://stackoverflow.com/a/59153244/5641227

jkowens:
[...]

PR #556, opened 2018-10-12
Add section about how to require activerecord-import

oniofchaos:
Closes https://github.com/zdennis/activerecord-import/issues/474

Replacement for https://github.com/zdennis/activerecord-import/wiki/Requiring

Part of https://github.com/zdennis/activerecord-import/issues/397

Follow up to https://github.com/zdennis/activerecord-import/pull/552

Issue #397, opened 2017-04-04
Improve Readme


jkowens:
The readme could use a makeover 💅 

Adding details for the following issues will be a good start. Pulling in some key information from the wiki would be a good idea too.

- [x] #174 
- [x] #229 
- [x] #233 
- [ ] #318 
- [x] #396  
- [x] #430 
- [x] #442 
- [x] #474 
- [x] #505 
- [x] #507

tibbon:
Hmm, I might take a swing at this when I get a chance.

jkowens:
@tibbon that would be fantastic!!

oniofchaos:
https://github.com/zdennis/activerecord-import/pull/556 checked off https://github.com/zdennis/activerecord-import/issues/474

oniofchaos:
As these get updated, how do you feel about replacing the content of the wikis with links to the appropriate README sections?

oniofchaos:
Also, for https://github.com/zdennis/activerecord-import/issues/396 the OC ended up closing out the issue because it went stale and he wasn't able to provide a demo since he doesn't work on that project anymore

jkowens:
Regarding the the wiki, that sounds good. Add links to the readme and eventually we will probably just delete the entries.

PR #563, opened 2018-11-02
Additional README updates

oniofchaos:
Followup to https://github.com/zdennis/activerecord-import/pull/560, https://github.com/zdennis/activerecord-import/pull/556, and, https://github.com/zdennis/activerecord-import/pull/554 in the series to address https://github.com/zdennis/activerecord-import/issues/397. 

Closes https://github.com/zdennis/activerecord-import/issues/430.
Closes https://github.com/zdennis/activerecord-import/issues/507. 

Migrates the following wiki articles into the README:
https://github.com/zdennis/activerecord-import/wiki
https://github.com/zdennis/activerecord-import/wiki/Counter-Cache-Column
https://github.com/zdennis/activerecord-import/wiki/Examples
https://github.com/zdennis/activerecord-import/wiki/How-to-run-the-tests
https://github.com/zdennis/activerecord-import/wiki/Note:-ActiveRecord-Timestamps
https://github.com/zdennis/activerecord-import/wiki/On-Duplicate-Key-Ignore
https://github.com/zdennis/activerecord-import/wiki/On-Duplicate-Key-Update
https://github.com/zdennis/activerecord-import/wiki/Supported-Database-Adapters

I did not add in the Benchmark Wiki as the results seemed old (Jan 2017), specific to a particular DB, and did not have the script used so could not be validated.

Since we now have the majority, if not everything, in the README, now would be a good time to review and read the document as a holistic item and suggest changes to it as such.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -21,12 +21,21 @@ and then the reviews:
 That would be about 4M SQL insert statements vs 3, which results in vastly improved performance. In our case, it converted
 an 18 hour batch process to <2 hrs.
 
+The gem provides the following high-level features:
+
+* activerecord-import can work with raw columns and arrays of values (fastest)
+* activerecord-import works with model objects (faster)
+* activerecord-import can perform validations (fast)
+* activerecord-import can perform on duplicate key updates (requires MySQL or Postgres 9.5+)
+
 ## Table of Contents
 
 * [Examples](#examples)
+  * [Introduction](#introduction)
   * [Columns and Arrays](#columns-and-arrays)
   * [ActiveRecord Models](#activerecord-models)
   * [Batching](#batching)
+  * [Recursive](#recursive)
 * [Options](#options)
   * [Duplicate Key Ignore](#duplicate-key-ignore)
   * [Duplicate Key Update](#duplicate-key-update)
@@ -47,6 +56,29 @@ an 18 hour batch process to <2 hrs.
   * [Running Tests](#running-tests)
 
 ### Examples
+
+#### Introduction
+
+Without `activerecord-import`, you'd write something like this:
+
+```ruby
+10.times do |i|
+  Book.create! :name => "book #{i}"
+end
+```
+
+This would end up making 10 SQL calls. YUCK!  With `activerecord-import`, you can instead do this:
+
+```ruby
+```ruby
+books = []
+10.times do |i|
+  books << Book.new(:name => "book #{i}")
+end
+Book.import books    # or use import!
+```
+
+and only have 1 SQL call. Much better!
 
 #### Columns and Arrays
 
@@ -127,7 +159,22 @@ columns = [ :title ]
 
 # 2 INSERT statements for 4 records
 Book.import columns, books, :batch_size => 2
+```
 
+#### Recursive
+
+NOTE: This only works with PostgreSQL.
+
+Assume that Books <code>has_many</code> Reviews.
+
+```ruby
+books = []
+10.times do |i|
+  book = Book.new(:name => "book #{i}")
+  book.reviews.build(:title => "Excellent")
+  books << book
+end
+Book.import books, recursive: true
 ```
 
 ### Options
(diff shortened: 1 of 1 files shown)


README.markdown
```diff
@@ -21,16 +21,313 @@ and then the reviews:
...
+
+By default, `activerecord-import` will not validate for uniquness when importing records. Starting with `v0.27.0`, there is a  parameter called `validate_uniqueness` that can be passed in to trigger this behavior. This option is provided with caution as there are many potential pitfalls. Please use with caution.
+
+```ruby
+Book.import books, validate_uniqueness: true
+```
+
+### Array of Hashes
+
+Due to the counter-intuitive behavior that can occur when dealing with hashes instead of ActiveRecord objects, `activerecord-import` will raise an exception when passed an array of hashes. If you have an array of hash attributes, you should instead use them to instantiate an array of ActiveRecord objects and then pass that into `import`.
```
oniofchaos: @jkowens @ahmad-sanad I want to confirm that I understood the issue here correctly.

README.markdown
```diff
@@ -179,6 +494,28 @@ For more information on activerecord-import please see its wiki: https://github.
 
 To document new information, please add to the README instead of the wiki. See https://github.com/zdennis/activerecord-import/issues/397 for discussion.
 
+### Contributing
```
oniofchaos: @jkowens not sure if there's any other info here you want to add

README.markdown
ahmad-sanad: So an array of hashes does work, but only if the columns are consistent in every hash in the array.
e.g.
```
MyModel.import [{foo: 1}, {foo: 2, bar: 3}] # raises an exception

MyModel.import [{foo: 1, bar: 2}, {foo: 3, bar: 4}] # works fine
```

The potential workarounds discussed were:

1. Convert it to an array of ActiveRecord objects
```
# Instead of this
MyModel.import [{foo: 1}, {foo: 2, bar: 3}]

# Do this
MyModel.import [MyModel.new(foo: 1), MyModel.new(foo: 2, bar: 3)]

# but hopefully with cleaner code lol
```

2. Divide the array into multiple ones with consistent columns and import each one separately
```
# Instead of this
MyModel.import [{foo: 1}, {foo: 2, bar: 3}] 

# Do this
MyModel.import [{foo: 1}] 
MyModel.import [{foo: 2, bar: 3}] 
```

ahmad-sanad: Also, thanks for adding this!!!

oniofchaos: NP! Thanks for the clarification. I will fix this up.

README.markdown
```diff
@@ -21,16 +21,360 @@ and then the reviews:
...
+
+# result in table books
+# title  | author
+#--------|--------
+# Book 1 | NULL
+# Book 2 | NULL
+
+```
+
+Using hashes will only work if the columns are consistent in every hash of the array. If this does not hold, an exception will be raised. There are two workarounds: use the array to instantiate an array of ActiveRecord objects and then pass that into `import` or divide the array into multiple ones with consistent columns and import each one separately.
```
oniofchaos: @jkowens @ahmad-sanad moved original answer to here

jkowens:
@oniofchaos thanks for all the work you've put into this documentation! I was under the weather last week, so I'm just getting caught up. I'll try to review soon and if all is good I'll get it merged in. 👍

README.markdown
jkowens: So in regards to no. 2 the thought was to make use of Enumerable's `#group_by`.

For example:

```
arr = [{foo: 1}, {foo: 2, bar: 3}] 
arr.group_by(&:keys).each_value do |v|
 MyModel.import v
end
```
I think this would be the recommended workaround, but I don't have any benchmark data to support that.

oniofchaos: Updated hash examples again with `group_by` @jkowens

jkowens:
Thanks again for putting this together, it gives us a great framework moving forward. I agree now that we have largely everything moved the the readme it will be easier to see where we can fill in more details or even make some things more concise.

oniofchaos:
I'm glad I was able to contribute to this gem. I use it all the time and it was nice to give back.

Issue #318, opened 2016-09-22
document import!


indigoviolet:
Seems like a useful method, but isn't mentioned anywhere (at least I didn't find it)

jkowens:
You're right, this should be added to the wiki!

soundasleep:
I added a quick note to https://github.com/zdennis/activerecord-import/wiki with it :)

oniofchaos:
It looks like this is pretty heavily documented in the code and in the RubyDoc code wiki. I'm not sure that anything more needs to be done here?

https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/import.rb
https://www.rubydoc.info/gems/activerecord-import/ActiveRecord%2FBase.bulk_import

oniofchaos:
Thoughts @jkowens ?

PR #572, opened 2018-11-29
Add options table for import to README

oniofchaos:
Closes https://github.com/zdennis/activerecord-import/issues/318
Closes https://github.com/zdennis/activerecord-import/issues/397

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -40,7 +40,6 @@ The gem provides the following high-level features:
 * [Options](#options)
   * [Duplicate Key Ignore](#duplicate-key-ignore)
   * [Duplicate Key Update](#duplicate-key-update)
-  * [Uniqueness Validation](#uniqueness-validation)
 * [Return Info](#return-info)
 * [Counter Cache](#counter-cache)
 * [ActiveRecord Timestamps](#activerecord-timestamps)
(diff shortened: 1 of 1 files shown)


README.markdown
```diff
@@ -247,10 +246,26 @@ Book.import books, recursive: true
 
 ### Options
 
+Key                     | Options               | Default            | Description
```
oniofchaos: I did my best scanning through the comments in the code to pick up all the options but double check me to make sure I didn't miss any options!

lib/activerecord-import/import.rb
```diff
@@ -285,32 +285,7 @@ def supports_setting_primary_key_of_imported_objects?
     # below for what +options+ are available.
     #
     # == Options
-    # * +validate+ - true|false, tells import whether or not to use
```
oniofchaos: Thoughts on this decision here? Also, thoughts on removing more of the documentation here as it's examples and we could link to the examples in the README?

jkowens: I don't think we want to remove the options from the comments here because they get picked up in the ruby doc:

https://www.rubydoc.info/github/zdennis/activerecord-import/ActiveRecord%2FBase.import

README.markdown
jkowens: Looks good!

lib/activerecord-import/import.rb
oniofchaos: That makes a lot of sense 👍

PR #584, opened 2019-01-23
Move ActiveRecord::Dirty changes to previous_changes after import

stokarenko:
There is a legal `ActiveRecord::Dirty.changes_applied` finalization method starting from `v4.1.8`.

It will clean the Dirty `changes` as usually, and will populate the `previous_changes` in additional - can be useful for further processing.

Actually, I need that right now :)

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -862,7 +862,9 @@ class ActiveRecord::Base
       end
 
       models.each do |model|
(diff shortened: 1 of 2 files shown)


stokarenko:
@jkowens Fixed! )

jkowens:
This has been released in v1.0.0 🎉

stokarenko:
many thanks! )

PR #589, opened 2019-02-01
A better error message for a corner case that resulted to hard-to-decipher errors.

golddranks:
Currently, when the user uses the arrays-of-arrays import mechanism, if there is more values than columns, it just produces broken SQL query such as
```
INSERT INTO table (col1, col2) VALUES ("val", "val2",)
```
Note the final comma. After it, there is a `nil` value that doesn't get printed. The reason the `nil` gets there is because if `column = columns[j]` in `values_sql_for_columns_and_attributes` overflows, `column` is going to be `nil` and not match any of the conditions, producing, again a `nil`.

A malformed SQL query is something definitely not desirable. I don't know about the general ideology behind this library, (whether you want to be loud about errors or ignore the case where there is too much data but you still manage to get on) but at least throwing an error message is certainly better than it currently is, and doesn't come here with an additional cost.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -976,6 +976,8 @@ class ActiveRecord::Base
               val = column.type_cast(val) unless column.type.to_sym == :binary
               connection_memo.quote(val, column)
             end
+          else
+            raise ArgumentError, "Too much values (#{arr.length} compared to columns #{columns.length})"
           end
         end
         "(#{my_values.join(',')})"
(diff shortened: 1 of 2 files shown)


golddranks:
Oh, just throw me a line if you want a test case but if it's just faster to implement it by yourselves and you have the time, I don't mind either way!

jkowens:
This is a really nice addition. If you can add a test to this PR I'll get it merged 🚀

golddranks:
Added the test. Merged accidentally, so I force-pushed a rebased version that keeps the history cleaner.

jkowens:
Nice work! 🎉

jkowens:
Released in v1.0.1!

PR #594, opened 2019-02-17
Fix import of saved json[b] fields for postgres

stokarenko:
Well, `activerecord-import` silently corrupted a tons of the data on my server 💣
Fortunately, that was a staging one, and it's possible to revert this time 🍀 😃 

How that happened:
1) I have a pure `jsonb` postgres field - without default value (don't ask me why 😞), without `serialize` nor  `store_accessor` definition on model layer.
2) I use `ActiveRecord.import` with `on_duplicate_key_update` turned on.
3) everything is fine when the new value comes there for `jsonb` field. But...
4) if `jsonb` field was not changed - `activerecord-import` will take it from `AR#read_attribute_before_type_cast`. As a result in DB it will replace the original `jsonb` object with that object serialized to JSON **twice**. Postgres will silently accepts that, because such **string** is valid JSON string...
5) Each time when import was called - the things becomes to be worse and worse for particular record 🔥, starting from normal `{"a": "b"}` JSON object, till `"\"\\\"{\\\\\\\"a\\\\\\\": \\\\\\\"b\\\\\\\"}\\\"\""` and further! 🚀

The fix is similar to https://github.com/zdennis/activerecord-import/commit/2a6158d68ff128c5c88a47d92b8b9051aff8e145. It fixes pure `json` case as well.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -577,10 +577,10 @@ class ActiveRecord::Base
         default_values = column_defaults
         stored_attrs = respond_to?(:stored_attributes) ? stored_attributes : {}
         serialized_attrs = if defined?(ActiveRecord::Type::Serialized)
-          attrs = column_names.select { |c|
+          attrs = column_names.select do |c|
             attribute_type = type_for_attribute(c.to_s)
-            attribute_type.class == ActiveRecord::Type::Serialized || %i[json jsonb].include?(attribute_type.type)
-          }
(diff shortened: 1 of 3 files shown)


stokarenko:
Travis failed due to the reason like https://github.com/travis-ci/travis-ci/issues/7987
Build rerun could help I believe

jkowens:
Looks like the Travis builds are working again. I'm just seeing some rubocop offenses now.

stokarenko:
@jkowens rubocop has been fixed! )

jkowens:
Looks good, thanks!

jkowens:
Released in v1.0.1!

PR #595, opened 2019-02-19
Don't pass returning on recursive import

dmitriy-kiriyenko:
Same reason `on_duplicate_key_update` is not passed to associations on
recursive import currently, must be applied to `returning` option, which
is otherwise unusable with recursive import.


jkowens:
Good catch!

jkowens:
Released in v1.0.1!

PR #605, opened 2019-03-18
add cockroachdb to adapter detection

willie:
Required for cockroachdb since it works with the postgresql adapter.

See https://www.cockroachlabs.com/docs/stable/build-a-ruby-app-with-cockroachdb-activerecord.html

### DIFF
--- a/lib/activerecord-import/base.rb
+++ b/lib/activerecord-import/base.rb
@@ -13,6 +13,7 @@ module ActiveRecord::Import
     when 'postgresql_makara' then 'postgresql'
     when 'makara_postgis' then 'postgresql'
     when 'postgis' then 'postgresql'
+    when 'cockroachdb' then 'postgresql'
     else adapter
     end
   end


lib/activerecord-import/base.rb
```diff
@@ -13,6 +13,7 @@ def self.base_adapter(adapter)
     when 'postgresql_makara' then 'postgresql'
     when 'makara_postgis' then 'postgresql'
     when 'postgis' then 'postgresql'
+    when 'cockroachdb' then 'postgresql'
```
oniofchaos: Looks like this needs to be added here too: https://github.com/zdennis/activerecord-import/blob/80bda221af3a9a478ae2c14c0edde5b210defa3e/Rakefile#L16

jkowens: @willie if you can make this small update I'll get this merged. Thanks!

jkowens: Actually, let's hold off on that for now.

jkowens:
Released in v1.0.2 🚀

PR #621, opened 2019-05-17
Update README: :validate is overridden by `import!`

Zeragamba:
Just got caught by this while testing AR-Import. the :validate option is overridden by the `import!` method. Here's an update to the Readme to make that a bit clearer.


jkowens:
Thanks!

Issue #635, opened 2019-08-07
Results for postgresql is nil when importing empty array


gyfis:
I know this is an ugly edge case, but when you `.import` an empty array, the resulting `results` for PostgreSQL is nil, where it probably should be an empty array (since `ids` is an empty array in the result):

```ruby
> import_result = RowTitle.import([:title], [], returning: :title)
=> #<struct ActiveRecord::Import::Result failed_instances=[], num_inserts=0, ids=[], results=nil>
> import_result.ids
=> []
> import_result.results
=> nil
```

It behaves the same when inputting an empty array without column titles into the `.import` method.

jkowens:
Yep, that should be consistent. We'd have to make that update part of a major release if is decided to fix.

gyfis:
@jkowens Cool! I prepared a PR: #636

PR #636, opened 2019-08-09
Return an empty array for PostgreSQL results instead of nil

gyfis:
Issue: #635


jkowens:
Thanks!

jkowens:
This has been released in v1.0.3 🎉

PR #642, opened 2019-10-01
Update for Rails edge/6.1 alpha

imtayadeway:
https://github.com/rails/rails/pull/37253 changed the connection pool
to pass in a database config instead of a specification - this updates
the base importer to be compatible with both


gregmolnar:
Looking good to me.

jkowens:
Thanks!

jkowens:
This has been released in v1.0.3 👍

PR #649, opened 2019-10-22
Update documentation to include description for :all_or_none option

dombesz:
The `:all_or_none` option was not documented and I made an edit to include it.


jkowens:
Thanks!

Issue #650, opened 2019-11-25
NoMethodError with PostgreSQL ENUM fields


sebcoetzee:
When using PostgreSQL `ENUM` fields, the import method raises a `NoMethodError: undefined method 'to_sym' for nil:NilClass`.

Given an enum type in PostgreSQL, defined as follows:

```sql
CREATE TYPE vendor_type AS ENUM ('wholesaler', 'retailer');
```
 and given that we create a field on the `vendor` table as follows:

```ruby
t.column :vendor_type, :vendor_type
```

then we get the `NoMethodError` when importing. Stepping through the code we see that on this line https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/import.rb#L967, the `column` value is as follows:

```
[1] pry(Vendor)> column
=> #<ActiveRecord::ConnectionAdapters::PostgreSQLColumn:0x00007fd9a25c2910
 @coder=nil,
 @default=nil,
 @limit=nil,
 @name="vendor_type",
 @null=true,
 @precision=nil,
 @primary=false,
 @scale=nil,
 @sql_type="vendor_type",
 @type=nil>
```

Since the `column.type` is `nil`, calling `column.type.to_sym` raises the `NoMethodError`.

Proposed solution:

Use the safe access operator to prevent a `NoMethodError` e.g.: `column.type&.to_sym`

PR #651, opened 2019-11-25
Fix NoMethodError when using PostgreSQL ENUM types

sebcoetzee:
This PR fixes the issue described in https://github.com/zdennis/activerecord-import/issues/650.

By checking if the `.type` is `nil`, there is no longer a risk of calling `.to_sym` on a `nil` value.

I have added a `ENUM` field to the PostgreSQL schema for the tests. Because PostgreSQL does not support using `IF NOT EXISTS` on a `CREATE TYPE` statement, I had to implement the same logic by rescuing the correct exception type and then calling `ALTER TYPE` in the `rescue` block.


jkowens:
Thanks for the contribution!

jkowens:
This has been released in v1.0.4 🎉

Issue #653, opened 2019-12-20
Document `validate_with_context` Option


jdalt:
The `README` should make light of the `validate_with_context` feature.

I struggled to find a way to avoid certain very expensive validations (that I could pre-validate en masse) and only found that I could pass a validation context when reading the source code. 

I think this feature should be called out in the options section: https://github.com/zdennis/activerecord-import#options

Thanks so much for creating and maintaining this gem! I've worked in Rails for several years and this solves a huge class of problems that I had never found a good solution for. This was the missing piece in my performance strategy, and I'm grateful to have found your work.

jkowens:
@jdalt thank you for the kind words! Feel free to send up a PR that adds documentation for this option. We'll get it merged in 👍

jdalt:
Thanks for the quick reply! I'm on it.

jdalt:
@jkowens https://github.com/zdennis/activerecord-import/pull/654

PR #654, opened 2019-12-20
Document :validate_with_context option in README

jdalt:
Resolves https://github.com/zdennis/activerecord-import/issues/653


jkowens:
Looks good, thanks!

PR #656, opened 2019-12-22
Fix #655: optimistic lock in Postgres

### LINKED ISSUE #655: If model uses AR optimistic locking, lock_version attribute is always reset to 1
This applies to at least Postgres implementation. 
**Repro:**

1. Create a model with `lock_version` numeric attribute, this will activate AR optimistic locking
2. Import with `import(... , on_duplicate_key_update: [])`
3. The generated SQL will look like: `INSERT INTO ... ON CONFLICT (...) DO UPDATE SET "updated_at"=EXCLUDED."updated_at","lock_version"=EXCLUDED."lock_version"+1  RETURNING ... `

Where `EXCLUDED."lock_version"` is 0 in the original fresh object. 

**Expected:**
Gererated SQL be `ON CONFLICT (...) DO UPDATE SET "updated_at"=EXCLUDED."updated_at","lock_version"="lock_version"+1` or even better 
`ON CONFLICT (...) DO UPDATE SET "updated_at"=EXCLUDED."updated_at","lock_version"=COALESCE("lock_version", 0)+1`

OR

Don't touch `lock_version` at all and let user do it with custom SQL. Now if you try to specify `lock_version` in `on_duplicate_key_update: []` array it will raise an error `PG::SyntaxError: ERROR:  multiple assignments to same column "lock_version"`

I think it is a bug/typo in https://github.com/zdennis/activerecord-import/pull/500 , namely here: https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/adapters/postgresql_adapter.rb#L208 it should refer to a `lock_version` in the table, not in inserted object.

timanovsky:
Fixes https://github.com/zdennis/activerecord-import/issues/655

For the record, the tricky case if `lock_version` is nullable and is null is handled neither by this gem, nor by AR (tested with v 4.2).

### DIFF
--- a/lib/activerecord-import/adapters/postgresql_adapter.rb
+++ b/lib/activerecord-import/adapters/postgresql_adapter.rb
@@ -158,7 +158,7 @@ module ActiveRecord::Import::PostgreSQLAdapter
       qc = quote_column_name( column )
       "#{qc}=EXCLUDED.#{qc}"
     end
-    increment_locking_column!(results, locking_column)
+    increment_locking_column!(table_name, results, locking_column)
     results.join( ',' )
   end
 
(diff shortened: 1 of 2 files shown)


timanovsky:
@jkowens Hi, are you planning to release any time soon?

jkowens:
This has been released in v1.0.4 🎉

PR #662, opened 2020-01-10
Allow Filtering of repeated instances when doing an import

estebanz01:
Sometimes, we might have repeated instances in the array passed to the import and those might cause a fail while inserting, because the same value is being inserted more than once when it shouldn't be possible. This is pretty useful for upsert operations.

This PR adds a `unique_records_by` option, which if present, depending on the value, it will try to filter the passed elements for import. It also works for hashes.


estebanz01:
CC @zdennis @jkowens in case notification of this PR didn't arrive :smile:

jkowens:
Thanks @estebanz01, I appreciate your interest in helping improve activerecord-import 👍  You certainly put a nice amount of effort into this PR. I'll have to review and think about this feature for a bit though. I'm getting more cautious about adding new options to the library, particularly if they can be handled fairly easily outside of the import call.

PR #648, opened 2019-10-14
Use prepend pattern for ActiveRecord::Base#establish_connection patching

dombesz:
This PR fixes the issue mentioned in the #647 .

PR #663, opened 2020-01-21
Fix compatibility issue because of not returning ActiveRecord::Connection from #establish_connection

reverentF:
After updating activerecord-import to v1.0.4, our project has started failing.

The cause of this problem is `#establish_connection` returns TrueClass/FalseClass (changed by #648).

Some gem expect the method to return `ActiveRecord::ConnectionAdapters::ConnectionPool` which is returned by original #establish_connection.
(I found this problem by using `activerecord-sharding` gem with activerecord-import.)

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -241,8 +241,9 @@ end
 
 module ActiveRecord::Import::Connection
   def establish_connection(args = nil)
-    super(args)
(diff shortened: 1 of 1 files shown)


jkowens:
Good catch, thanks! 👍

jkowens:
This has been released in v1.0.5.

PR #671, opened 2020-04-04
when having a custom `.to_sql` method, import fails due to syntax errors

florianeck:
in my projects, `DateTime` responds to `.to_sql` for working with them in DB context... this causes `activerecord-import` to fail on datetime fields....

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -961,7 +961,7 @@ class ActiveRecord::Base
           if val.nil? && Array(primary_key).first == column.name && !sequence_name.blank?
             connection_memo.next_value_for_sequence(sequence_name)
           elsif val.respond_to?(:to_sql)
-            "(#{val.to_sql})"
+            "('#{val.to_sql}')"
           elsif column
             if respond_to?(:type_caster)                                         # Rails 5.0 and higher
               type = type_for_attribute(column.name)


lib/activerecord-import/import.rb
```diff
@@ -961,7 +961,7 @@ def values_sql_for_columns_and_attributes(columns, array_of_attributes) # :nodoc
           if val.nil? && Array(primary_key).first == column.name && !sequence_name.blank?
             connection_memo.next_value_for_sequence(sequence_name)
           elsif val.respond_to?(:to_sql)
-            "(#{val.to_sql})"
+            "('#{val.to_sql}')"
```
jkowens, line 964: This isn't going to work. The `to_sql` call can return sub queries.

jkowens:
Can you share an example of what a failing sql insert statement looks like in your project?

lib/activerecord-import/import.rb
florianeck: i see..

florianeck:
@jkowens i mostly get the issue with dates... i found a workaround for by doing sth like this:

instead:

`items_to_insert << Item.new(name: 'foo', date: Date.new(2020,1,1)) # creates problems when inserting`

is do this:

`items_to_insert << { name: 'foo', date: Date.new(2020,1,1).to_s } # works fine`

jkowens:
Nice, that's great to hear you have an easy workaround. 👍

PR #673, opened 2020-04-22
Skip load association ids when column changed

Aristat:
We define association id even when we change this column, because of this it is impossible to set `nil`
Maybe we can skip such `load associations`, what do you think?


jkowens:
I think this looks good, thanks!

hathawad:
@jkowens is it possible to get a release w/ this change in it? Thanks!

jkowens:
This has been released in v1.0.5 👍

PR #674, opened 2020-04-22
Update README to note early on that `import` might be named `bulk_import` for compatibility reasons

lvreynoso:
### Description
Adds a line in the introduction of the README that notes the `import` method might not be aliased as that, and would have to be called with `bulk_import` if other conflicting gems are present. Since the method is a such major and important part of this gem, I thought it would be best to quickly note this upfront since the debugging information encountered might be a bit mysterious.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -60,6 +60,8 @@ The gem provides the following high-level features:
 
 #### Introduction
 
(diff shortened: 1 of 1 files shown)


jkowens:
Looks good, thanks!

Issue #597, opened 2019-02-22
Raise/warn when trying to import hash recursively (or add documentation?)


dgollahon:
Hi,

I found that recursive import with hashes is not supported (https://github.com/zdennis/activerecord-import/issues/522#issuecomment-397033165) and this is ok, but it took me a little while to realize this after getting a fairly opaque error message. I think it would be an improvement if `activerecord-import` raised or warned on the combination of `recursive: true` and an array of hashes or, alternatively, if a warning about this appeared in the [hashes](https://github.com/zdennis/activerecord-import#hashes) section of the readme.

Thanks!

jkowens:
I think a combination of a warning message and more documentation would be great 👍

PR #677, opened 2020-05-17
Document that hashes and arrays won't work in recursive mode.

javidh:
This can be [confusing](https://github.com/zdennis/activerecord-import/issues/597), so this can help.


jkowens:
Looks good, thanks!

PR #684, opened 2020-07-01
Provide a mechanism to map failures back to input

rorymckinley:
This change adds functionality that makes it easy to tell a user which of records in a dataset have failed validation. We found this to be particularly useful when dealing with imports from spreadsheets (e.g. `the record in line x has the following errors....`).

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -541,7 +541,7 @@ class ActiveRecord::Base
     alias import! bulk_import! unless ActiveRecord::Base.respond_to? :import!
 
     def import_helper( *args )
-      options = { validate: true, timestamps: true }
+      options = { validate: true, timestamps: true, track_validation_failures: false }
       options.merge!( args.pop ) if args.last.is_a? Hash
       # making sure that current model's primary key is used
       options[:primary_key] = primary_key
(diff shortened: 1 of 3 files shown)


lib/activerecord-import/import.rb
```diff
@@ -704,7 +704,7 @@ def import_helper( *args )
               array_of_attributes[i] = nil
               failure = model.dup
               failure.errors.send(:initialize_dup, model.errors)
-              failed_instances << failure
+              failed_instances << (options[:track_validation_failures] ? [i, failure] : failure )
```
jkowens, line 707: Are there scenarios where `i` doesn't match the index of the record in the `failed_instances` array?

rorymckinley: Apologies for the slow response. As `i` is generated by `arr.each_with_index do |hsh, i|` it should match the index of the record in the original dataset but it will not match the failure's position in `failed_instances`, except for the sad case where all the records have failed.

`arr` maintains the sequence of `array_of_attributes` and, as far as I can tell, `array_of_attributes` maintains the sequence of the original inputs.

jkowens: Ah yes, ok that makes sense.

jkowens:
Looks good, thanks for your contribution!

jkowens:
This has been released in v1.0.7. Thanks!

Minoru-Noda (commented):
Sorry, I accidentally reviewed this pull request.

PR #691, opened 2020-07-25
use uninitialized object to get rails timestamps

AhMohsen46:


jkowens:
Looks good 👍

AhMohsen46:
Thanks @jkowens, glad it helped :D

PR #699, opened 2020-08-07
Fix mysql schema file name in benchmarks

Slike9:
The PR fixes running command
```
ruby benchmark.rb --adapter=mysql2 --to-html=results.html
```
The command failed with
```
Mysql2::Error: Table 'activerecord_import_test.test_myisam' doesn't exist
```
Because the schema file name was incorrect, the schema was not loaded
and the tables were not created.


jkowens:
Thanks 👍

PR #707, opened 2020-09-25
GitHub is HTTPS by default

amatsuda:
This reduces unnecessary redirection from http://... to https://... when someone visits this gem's homepage via https://rubygems.org/gems/activerecord-import or some tools or APIs that use the gem's metadata.


jkowens:
Thanks!

PR #708, opened 2020-09-25
Fetch value from correct field/col from result row

diclophis:
@jkowens I sincerely apologize for this ... your sorta scooped up my PR faster than I was able to process it through our QA process for the application we are integrating this gem into... I think the result row has essentially only 1 column now given the new `SELECT @@` syntax... 

I would really like to backfill some spec/test code here, but I am having a really hard time finding the correct strategy for testing this code.

Site note: see also the rails implementation of `show_variable` https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/AbstractMysqlAdapter.html#method-i-show_variable (the @@ syntax is what they have been using since at least rails 4)


jkowens:
No worries, I thought we would have had existing test that would have covered that...but apparently not.

jkowens:
The probably found it was difficult to test as well 😄

Issue #690, opened 2020-07-24
how to handle `after_initialize`?


breathe:
I have a rails model with an `after_initialize` method that enforces some non-nil fields as early as possible in the lifecycle of the program ...

```
class GqlLocationInfo < ActiveRecord::Base

  # ugh -- field named 'type' is special in activerecord ...
  # disable that behavior so that field behaves as regular column
  self.inheritance_column = nil

  after_initialize :new_record_defaults, :if => :new_record?

  private

  def new_record_defaults
    if self.type.nil?
      raise ArgumentError, "Invariant violation: type must be provided"
    end

    if self.revision.nil?
      raise ArgumentError, "Invariant violation: revision must be provided"
    end

    if self.location_id.nil?
      raise ArgumentError, "Invariant violation: location_id must be provided"
    end

    self.zone = GqlLocation::zone

    if self.zone.nil?
      raise ArgumentError, "Invariant violation: zone cannot be nil"
    end

    self.version = GqlLocation::current_storage_version(for_type: self.type)

    if self.version.nil?
      raise ArgumentError, "Invariant violation: version cannot be nil"
    end
  end

```

Is there anyway to use such a model with activerecord-import?

I've tried with ar records and hashes and I always get failures inside of new_record_defaults -- there is some point within active-record-import where the model above is instantianted with no properties set at all ...

AhMohsen46:
after_initialize was being called on initializing objects in [import.rb#L998](https://github.com/zdennis/activerecord-import/blob/c7660a488ea76fb55182b3ce05b514fd357c866c/lib/activerecord-import/import.rb#L998) and [import.rb#L579](https://github.com/zdennis/activerecord-import/blob/c7660a488ea76fb55182b3ce05b514fd357c866c/lib/activerecord-import/import.rb#L579)
as calling Model.new was initializing a new object with no attributes, and since you were validating on attributes presence, an error was being raised
to solve this I've created this [PR#691](https://github.com/zdennis/activerecord-import/pull/691)
where I've used [allocate](https://apidock.com/ruby/Class/allocate) instead of [new](https://apidock.com/ruby/Class/new)
the difference is that `new` was calls `allocate` then calls initialize, since `initialize` doesn't serve the purpose of fetching the timestamps_attributes and runs the `after_initialize` callback, which causes the error, the PR uses allocate instead for this purpose

jkowens:
@AhMohsen46 nice work!

AhMohsen46:
[...]

PR #709, opened 2020-09-29
to prevent mass-assignment errors in strict mode

diclophis:
We run our application in `:strict` mass assignment error mode in our test suite, and make heavy use of the this gem ... I would like to propose switching back to the longer form of attribute assignment to prevent errors with some of our models.

If we could open a discussion on this PR it would be very helpful for us to understand why the change was made, and what options we have going forward.

It looks like the short-form of assignment was introduced here: https://github.com/zdennis/activerecord-import/compare/v1.0.5...v1.0.6#diff-70d0c2152cc72037bcf5c177f03eb9f8L700-L701

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -703,7 +703,8 @@ class ActiveRecord::Base
             # keep track of the instance and the position it is currently at. if this fails
             # validation we'll use the index to remove it from the array_of_attributes
             arr.each_with_index do |hsh, i|
-              model = new(hsh)
+              model = new
+              hsh.each_pair { |k, v| model[k] = v }
               next if validator.valid_model?(model)
               raise(ActiveRecord::RecordInvalid, model) if options[:raise_error]
               array_of_attributes[i] = nil
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -703,10 +703,13 @@ class ActiveRecord::Base
             # keep track of the instance and the position it is currently at. if this fails
             # validation we'll use the index to remove it from the array_of_attributes
             arr.each_with_index do |hsh, i|
-              model = new
-              hsh.each_pair { |k, v| model[k] = v }
+              model = new do |m|
+                hsh.each_pair { |k, v| m[k] = v }
+              end
+
               next if validator.valid_model?(model)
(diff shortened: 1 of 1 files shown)


jkowens:
Here is the history of that change:

https://github.com/zdennis/activerecord-import/issues/690

diclophis:
Ok great, let me dig into the referenced PR, and the test failures (which is good news that the change caused a regression)

lib/activerecord-import/import.rb
```diff
@@ -703,7 +703,8 @@ def import_helper( *args )
             # keep track of the instance and the position it is currently at. if this fails
             # validation we'll use the index to remove it from the array_of_attributes
             arr.each_with_index do |hsh, i|
-              model = new(hsh)
+              model = new
+              hsh.each_pair { |k, v| model[k] = v }
```
jkowens, line 707: I wonder if passing a block to new would solve this issue?

```
model = new do |m|
  hsh.each_pair { |k, v| m[k] = v }
end
```

jkowens:
@diclophis does this solve the issue running in :strict mass assignment error mode?

diclophis:
heya @jkowens I am still running through some QA on our app side, I will report back by EOD hopefully with some results.

lib/activerecord-import/import.rb
```diff
@@ -703,9 +703,13 @@ def import_helper( *args )
             # keep track of the instance and the position it is currently at. if this fails
             # validation we'll use the index to remove it from the array_of_attributes
             arr.each_with_index do |hsh, i|
-              model = new(hsh)
+              model = new do |m|
```
jkowens, line 706: @diclophis if this resolves your issue, can you put a comment here why we needed to pass a block to new?

diclophis: will do, i am just now getting through my test suite for this change

diclophis:
@zdennis I was able to cross check this, it looks OK on our side, is there any plan to publish a new gem version with some of the recent changes?

diclophis:
@jkowens sorry I also meant to ping you on this as well...

jkowens:
Thanks @diclophis,  I should be able to cut a new release this week 👍

jkowens:
This has been released in v1.0.7. Thanks!

Issue #715, opened 2020-12-02
Using activerecord-import with Rails 6.1.0.rc1 results in SystemStackError: stack level too deep


codeodor:
Using activerecord-import with Rails 6.1.0.rc1 results in SystemStackError: stack level too deep any time we attempt to establish our own connection. 

The line in question is: 

```
activerecord-import-1.0.7/lib/activerecord-import/import.rb:250:in `establish_connection'
```

codeodor:
It doesn't seem to happen if I just run one single test, but if I run my whole suite it always ends up triggering it. 

If I comment out the line `prepend ActiveRecord::Import::Connection` that adds the AR Import gem's `establish_connection` then it no longer triggers the error. (I don't yet know if AR Import works without doing that, though).

jkowens:
Hey @codeodor, thanks for the report. I had a feeling we would see some issues with Rails 6.1 since it sounds like there have some big updates in ActiveRecord. I think commenting out that prepend will basically disable AR Import. If you happen upon any leads please keep us updated!

codeodor:
Thanks @jkowens. In the one place I'm using AR Import, it still seems to work when commenting out that line. That's the good news! But, I'm not sure why AR Import has overridden AR::Base connection method? 

It looks like it needs to run this: 

`ActiveRecord::Import.load_from_connection_pool connection_pool`

But does it need to run it on every AR::Base model, and what does that do? 

I traced it back to 10 years ago: https://github.com/zdennis/activerecord-import/blame/3c08fad1a2cfa9d7880730f068ddb318aa7b2968/lib/activerecord-import.rb

But could it not just get the connection and do that inside the import method? (And is that necessary, even?) 

Those are earnest questions. I don't know it well enough, so I'm just throwing things out there. :)

codeodor:
When running the tests on mysql2 with that prepend removed, we do get more errors, most of which seem to be of the sort:

`ActiveRecord::RecordNotUnique: Mysql2::Error: Duplicate entry '99' for key 'PRIMARY'`

Strangely, sqlite3 tests work better with that line removed: only 4 failures and 3 errors vs 0 failures and 138 errors with it on.

codeodor:
If we move that line to the bulk_import method, all the tests seem to go back to passing (the ones that passed to begin with, before any changes were made, anyway). 

I will check if it gets rid of the problem in my app like this.

jkowens:
[...]

PR #716, opened 2020-12-02
Only load connection pool for the one adapter that needs it, rather than on every connection

codeodor:
What do you think vis-a-vis #715? Any reason we couldn't do this connection pool load at the time of import rather than at load time for every AR model?

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -255,7 +255,7 @@ end
 
 class ActiveRecord::Base
   class << self
-    prepend ActiveRecord::Import::Connection
+    prepend ActiveRecord::Import::Connection if defined?(SeamlessDatabasePool)
 
     # Returns true if the current database connection adapter
     # supports import functionality, otherwise returns false.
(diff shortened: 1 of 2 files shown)


jkowens:
I'm not ignoring this, just haven't had a chance to dig into the failing tests yet 😅

codeodor:
Ok, I have tried some more and I just cannot find a simple enough example to reproduce this. However, moving this does fix my error and still allows my imports to run correctly, and the tests all seem to pass (locally anyway, I am posting this before the Travis CI build finishes). 

Any chance it could still be merged? 

If not, is there another approach that could be used to do this connection pool trick without prepending AR::Base?

jkowens:
The only issue I see is that it seems to break compatibility with the seamless_database_pool adapter. I don't know much about that one, so that will require some investigation.

codeodor:
Ok. I guess that fixes it. My apps tests now run without the stack over flow errors, and these tests pass. I guess only that one adapter needed that code, so if we only run it for that adapter, it will at least fix the rest of the issue. 

I'll try to clean it up into one commit tomorrow. 

What do you think?

codeodor:
@jkowens I think it's good now, if you can take a look when you have a moment! Thanks!!

jkowens:
@codeodor I'm going to close this for now, sorry it took a while to track down the root cause of the issue.

PR #719, opened 2021-02-01
Better query cache clearing strategy

EtienneDepaulis:
This PR is an attempt to fix an issue we experienced when migrating our rails app from 6.0 to 6.1.

We recently tried to migrate our Rails app to the latest version (6.1). Our test suites being green we decided to upgrade our production environment. That's when we started to see a random error:

`can't add a new key into hash during iteration` in `lib/active_record/connection_adapters/abstract/query_cache.rb`

This issue occurs in the `cache_sql` method: https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb#L115

Not all our rails app have this issue so it's definitively linked to a gem we are using. When searching in all our gems only one seems to manipulate the `@query_cache` object.

In your PG adapter, you use `query_cache.clear`. In the `QueryCache` class they have a different method which seems to be using a lock : https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb#L90

I wasn't able to reproduce the bug with a test (concurrency issues are pretty hard to emulate).

The other thing bugging me is that this caching behaviour has been there for ages, I clearly have no explanations about why it started failing with 6.1+ :(

### DIFF
--- a/lib/activerecord-import/adapters/postgresql_adapter.rb
+++ b/lib/activerecord-import/adapters/postgresql_adapter.rb
@@ -28,7 +28,7 @@ module ActiveRecord::Import::PostgreSQLAdapter
       else
         select_values( sql2insert, *args )
       end
-      query_cache.clear_query_cache if query_cache_enabled
+      clear_query_cache if query_cache_enabled
     end
 
     if options[:returning].blank?
--- a/lib/activerecord-import/adapters/postgresql_adapter.rb
+++ b/lib/activerecord-import/adapters/postgresql_adapter.rb
@@ -28,7 +28,7 @@ module ActiveRecord::Import::PostgreSQLAdapter
       else
         select_values( sql2insert, *args )
       end
-      query_cache.clear if query_cache_enabled
+      query_cache.clear_query_cache if query_cache_enabled
     end
 
     if options[:returning].blank?


jkowens:
Thanks for the contribution 👍   Are you able to confirm this fixes the issue you were experiencing?

EtienneDepaulis:
@jkowens not yet, I first wanted to present the idea. 
As this bug happens only under heavy load, it is hard to simulate.
I will try this new version this week and update this discussion with my findings.
Thanks for merging so quickly BTW :)

zachzillion:
@EtienneDepaulis Were you able to confirm whether this fixed the issue for you? We are running into the same thing

EtienneDepaulis:
@zachzillion it didn't fix the issue, we are still digging sadly :(

jkowens:
I released this update in v1.0.8 anyway. It looked like the right thing to do 🙃

davidrees:
@EtienneDepaulis did you have any luck tracking down the error?

EtienneDepaulis:
@davidrees We think the issue came from the newrelic gem <8.0
We don't have any issue anymore

davidrees:
@EtienneDepaulis Thank you!

Issue #723, opened 2021-03-12
ActiveModelImportResult[:num_inserts]  does not work In mysql


tashirosota:
Showld not num_inserts return if does not work, it gets confusing.

tashirosota:
What do you think of [this approach?](https://github.com/tashirosota/activerecord-import/pull/2/files)

jkowens:
Hi @tashirosota, `num_inserts` should be working for MySQL. If it isn't I believe that is a bug to be fixed. Can you post an example with result?

tashirosota:
@jkowens 
Sorry I was misunderstanding.
I thought num_inserts means record increased count.
This mistake is easy to happen.In fact our other team member made the same mistake.
How about adding warning log like [this](https://github.com/zdennis/activerecord-import/pull/724)?

PR #724, opened 2021-03-13
Add warning to ActiveRecord::Import::Result unsupported attribute by other than postgresql.

tashirosota:
ActiveRecord::Import::Result has some unsupported attributes by other than postgresql.
So I think this attribute must out warning if called by other than postgresql. 
issues: https://github.com/zdennis/activerecord-import/issues/723

### DIFF
--- a/lib/activerecord-import/import/result.rb
+++ b/lib/activerecord-import/import/result.rb
@@ -1,0 +1,37 @@
+module ActiveRecord::Import #:nodoc:
+  class Result
+    attr_accessor :failed_instances
+    attr_writer :num_inserts, :ids, :results
+    def initialize(failed_instances, num_inserts, ids, results)
+      @failed_instances = failed_instances
+      @num_inserts = num_inserts
+      @ids = ids
+      @results = results
+    end
+
+    def num_inserts(with_warn: true)
+      warn 'Info: ActiveRecord::Import::Result#num_inserts is not an increased record count. It is the number of inserts it took to import the data.' if with_warn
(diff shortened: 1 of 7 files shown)


tashirosota:
Hi @jkowens.
What do you think this?

lib/activerecord-import/import/result.rb
```diff
@@ -0,0 +1,37 @@
...
+    attr_writer :num_inserts, :ids, :results
+    def initialize(failed_instances, num_inserts, ids, results)
+      @failed_instances = failed_instances
+      @num_inserts = num_inserts
+      @ids = ids
+      @results = results
+    end
+
+    def num_inserts(with_warn: true)
+      warn 'Info: ActiveRecord::Import::Result#num_inserts is not an increased record count. It is the number of inserts it took to import the data.' if with_warn
```
tashirosota, line 13: This mistake is easy to happen.In fact our other team member made the same mistake.

jkowens: I don't think I like this warning here. I know it's confusing, but this seems to get in the way a bit.

tashirosota: @jkowens 
Thanks for review.
OK. I'll think about other approach.

jkowens: @tashirosota I'm good with the warning for `ids` and `results` though. If you just want to remove the warning for `num_inserts` I'm ok with moving forward with this.

PR #732, opened 2021-05-27
fix tiny typos

shutooike:
I found tiny typos, so fixed them!


jkowens:
Looks good, thank you!

PR #737, opened 2021-07-06
Compatibility with Composite Primary Key Gem

dlanileonardo:


jkowens:
Thanks for the submission. I just realized we need to get Github Actions configured to run the test suite now that Travis is unavailable.

jkowens:
Looks good! Thanks again 🙌

jkowens:
Released in v1.2.0 🎉

PR #741, opened 2021-07-13
Use polymorphic_name instead of base_class.name for imports

kmhajjar:
This will enable compatibility with activerecord-import when overriding polymorphic_name in an active record model. 

polymorphic name defaults to the base_class name so this should not have any impact:
https://github.com/rails/rails/pull/32148/files#diff-89b75a2bbbb593fbb0f1b902769c2001fddb0fc9b710048842549e14a03b0028R159.


kmhajjar:
@zdennis let me know what you think! Thanks!

jkowens:
@kmhajjar this looks great, any chance you could update this PR so we can continue to support back to Rails 4.2?

kmhajjar:
> @kmhajjar this looks great, any chance you could update this PR so we can continue to support back to Rails 4.2?

Awesome thanks, was about to push up a test and a fix for that! @jkowens would you mind creating a new release/version after it is approved?

kmhajjar:
> > @kmhajjar this looks great, any chance you could update this PR so we can continue to support back to Rails 4.2?
> 
> Awesome thanks, was about to push up a test and a fix for that! @jkowens would you mind creating a new release/version after it is approved?

Fix pushed

lib/activerecord-import/import.rb
```diff
@@ -961,8 +961,9 @@ def find_associated_objects_for_import(associated_objects_by_class, model)
         changed_objects.each do |child|
           child.public_send("#{association_reflection.foreign_key}=", model.id)
           # For polymorphic associations
+
           association_reflection.type.try do |type|
-            child.public_send("#{type}=", model.class.base_class.name)
+            child.public_send("#{type}=", model.class&.polymorphic_name || model.class.base_class.name)
```
jkowens, line 966: I think you might need to use `respond_to?` to determine what call to make.

kmhajjar: oh great catch! Thank you

lib/activerecord-import/import.rb
```diff
@@ -961,8 +961,12 @@ def find_associated_objects_for_import(associated_objects_by_class, model)
         changed_objects.each do |child|
           child.public_send("#{association_reflection.foreign_key}=", model.id)
           # For polymorphic associations
+          association_name = model.class.base_class.name
+          if model.class.respond_to?(:polymorphic_name)
+            model.class.polymorphic_name
```
jkowens, line 966: You can assign `association_name` here. Or you could do something like:

```
association_name = if model.class.respond_to?(:polymorphic_name)
    ...
else
    ...
end

jkowens:
Great work, thanks!

jkowens:
I will definitely cut a release soon, I was hoping to get a couple of the other recently opened issues addressed if possible.

kmhajjar:
@jkowens thank you so much!! Any chance this could get released by EOW? Or do you think it'll take a bit longer than that?

jkowens:
I'll try to get to it this weekend, no promises tho 😅

kmhajjar:
No prob!! Thanks so much :)

jkowens:
Released in v1.2.0 🎉

kmhajjar:
> Released in v1.2.0 🎉

Awesome thank you so much!!

PR #746, opened 2021-08-31
Support :recursive in combination with on_duplicate_key_update: :all.

deathwish:
This handles the common use case of "I want to upsert an entire tree of models, replacing all the old content" without changing the options format.


jkowens:
Cool, this seems reasonable. Thanks 👍

jkowens:

PR #750, opened 2021-09-21
Add support for compound foreign keys

Uladzimiro:
We use partitioning and for this, we need to include a partition key into each request. Hence, we need to redefine simple `belongs_to`, `has_many` to have (foreign_key, partition_key) as association connection. For example, if we partition tables by `account_id` and we have two tables `customers` and `orders`:

```ruby
# == Schema Information
#
# Table name: customers
#
#  id          :integer          not null, primary key
#  account_id  :integer
#  name        :string

class Customer < ActiveRecord::Base
  has_many :orders,
           inverse_of: :customer,
           primary_key: %i[account_id id],
           foreign_key: %i[account_id customer_id]
end

# == Schema Information
#
# Table name: orders
#
#  id           :integer          not null, primary key
#  account_id   :integer
#  customer_id  :integer
#  amount       :integer

class Order < ActiveRecord::Base
  belongs_to :customer,
             inverse_of: :orders,
             primary_key: %i[account_id id],
             foreign_key: %i[account_id customer_id]
end
```

We will get account_id as part the generated SQL when associations are called:

```ruby
customer = Customer.where(account_id: 1).first
customer.orders
=> SELECT * FROM orders WHERE account_id = 1 AND customer_id = 10;
```

However, the import of orders using `activerecord-import` will fail on `load_association_ids` method.
This PR attempts to fix the issue by treating all associations as a compound association. To simplify/unify the code we convert a single key association to an array or association keys and then work with each individual key the same way we worked before.

### DIFF
--- a/test/support/postgresql/import_examples.rb
+++ b/test/support/postgresql/import_examples.rb
@@ -78,21 +78,25 @@ def should_support_postgresql_import_functionality
       end
     end
 
-    describe "with compound foreign keys" do
-      let(:account_id) { 555 }
-      let(:customer) { Customer.new(account_id: account_id, name: "foo") }
-      let(:order) { Order.new(account_id: account_id, amount: 100, customer: customer) }
+    unless ENV["SKIP_COMPOSITE_PK"]
+      describe "with compound foreign keys" do
+        let(:account_id) { 555 }
+        let(:customer) { Customer.new(account_id: account_id, name: "foo") }
+        let(:order) { Order.new(account_id: account_id, amount: 100, customer: customer) }
 
-      it "imports and correctly maps foreign keys" do
-        assert_difference "Customer.count", +1 do
-          Customer.import [customer]
+        it "imports and correctly maps foreign keys" do
+          assert_difference "Customer.count", +1 do
+            Customer.import [customer]
+          end
+
+          assert_difference "Order.count", +1 do
+            Order.import [order]
+          end
+
+          db_customer = Customer.last
+          db_order = Order.last
+          assert_equal db_customer.orders.last, db_order
         end
-
-        assert_difference "Order.count", +1 do
-          Order.import [order]
-        end
-
-        assert_equal Customer.last.orders.last, Order.last
       end
     end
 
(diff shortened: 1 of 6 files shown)


Uladzimiro:
Thanks! Yes, I was looking into it. Will it be OK if I add two new separate models (Customer and Order) just to test this case?

jkowens:
Yes, that's no problem 👍

test/support/postgresql/import_examples.rb
```diff
@@ -78,6 +78,24 @@ def should_support_postgresql_import_functionality
       end
     end
 
+    describe "with compound foreign keys" do
```
jkowens, line 81: You'll want to wrap this in a `unless ENV["SKIP_COMPOSITE_PK"]` condition since composite_primary_keys gem doesn't support Rails 7 yet.

Uladzimiro: Oh, thank you! I was trying to understand what's wrong with Rails 7!

Uladzimiro:
@jkowens should I add the same test to MySQL and SQLite import examples, or adding PostgreSQL is enough?

jkowens:
It’s probably good enough, but could that test be moved to the import test file that is run by all three adapters?

jkowens:
@Uladzimiro I think you test could be moved here: https://github.com/zdennis/activerecord-import/blob/dcd602e6040bcb581b97b7c88e4cd638586cdfa9/test/import_test.rb#L144

Then it would run against the three different db adapters.

Uladzimiro:
@jkowens look, now we have MySQL test failing because it doesn't support `supports_setting_primary_key_of_imported_objects?` https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/import.rb#L592

But I guess SQLite supports this functionality. What would be the best way to write a test in a clean way? Copy tests to PostgreSQL and SQLite-specific files?

jkowens:
Ahh that’s right. Ok nevermind me, this can just stay in the Postgres test file.

Uladzimiro:
OK, then I'm removing this ENV condition and moving the test to postgresql specific file, right?

jkowens:
Yeah that's totally fine, sorry for the run around.

jkowens:
Great, thank you 🎉

jkowens:

PR #753, opened 2021-11-08
ActiveRecord `#database_version` is a public method

beauraF:
Hello,

`#database_version` is a public method in Rails. By overriding it, and making it private `activerecord-import` have side effects on some Rails features. Like structure dump since https://github.com/rails/rails/pull/43216.

```
NoMethodError: private method `database_version' called for #<ActiveRecord::ConnectionAdapters::PostgreSQLAdapter:0x00007f8d71dc7480>
```

In this PR I make `database_version` public again.

Have a good day ☀️


jkowens:
Thanks!

jkowens:

PR #760, opened 2021-12-31
enable frozen string literals

desheikh:
Enable compatibility for frozen string literals for newer rails versions.
This allows apps to be run with `RUBYOPT="--enable-frozen-string-literal"`


jkowens:
Looks good, thanks!

PR #752, opened 2021-10-27
fix Rails 7 deprecation with default_timezone

ryanwood:
Fixes #751

PR #762, opened 2022-01-13
Call `ActiveRecord.default_timezone=` if possible

mishina2228:
This PR is a follow-up to #752.

The following deprecation warning is displayed when I run CI with Rails 7.0.
```
DEPRECATION WARNING: ActiveRecord::Base.default_timezone= is deprecated and will be removed in Rails 7.1.
Use `ActiveRecord.default_timezone=` instead.
 (called from <top (required)> at /home/runner/work/activerecord-import/activerecord-import/test/test_helper.rb:61)
```
This PR will suppress it.


jkowens:
Thanks!

PR #763, opened 2022-01-16
Refactor CI

mishina2228:
* Remove consideration of Ruby 1.9 and 2.3 which are not supported 
* Enable ruby/setup-ruby's `bundler-cache` to use caching
* Split test and lint, and run RuboCop only once


jkowens:
Thank you!

PR #764, opened 2022-01-17
CI against Ruby 3.0 and 3.1

mishina2228:
This PR adds the Ruby 3.0 and 3.1 to the CI matrix, and does the following:

* Split tests by DB to allow for skipping.
* Since [seamless_database_pool](https://github.com/bdurand/seamless_database_pool) requires Ruby ~> 2.0, skip installing it and running tests with it for Ruby 3.0+.
* Use `YAML.safe_load_file` to allow the use of aliases with Ruby 3.1.
* Remove `rb-readline` from Gemfile to suppress a bunch of warnings.
And I confirmed all CIs are green.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -62,15 +62,21 @@ jobs:
           cp test/github/database.yml test/database.yml
         env:
           PGPASSWORD: postgres
-      - name: Run tests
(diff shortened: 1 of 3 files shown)


jkowens:
Nice work!

PR #769, opened 2022-05-17
Allow raw SQL in returning option

leonidkroka:


jkowens:
@leonidkroka this is really cool. Could you resolve the failures for ActiveRecord 4.2? It seems like it doesn't deserialize the boolean value for some reason.

leonidkroka:
@jkowens Fixed tests by adding a switcher for boolean values dependent on AR version.

---

Indeed, ActiveRecord 4.2 doesn't deserialize boolean values by default, and [manual deserialization](https://github.com/zdennis/activerecord-import/blob/master/lib/activerecord-import/import.rb#L867) takes care of the only fields, which are defined in the schema. 

Also, it seems a similar issue presents for further AR versions and other types(e.g. decimal `RETURNING "discount"`):
```
AR_VERSION=6.0
Promotion(promotion_id: integer, discount: decimal)

promotion = Promotion.new(discount: 0.1)
result = Promotion.import([promotion], returning: :discount)

result.results
=> ["0.1"]
promotion.discount
=> 0.1e0
```

Although pg adapter returns column types(`result.column_types`) for columns, which weren't deserialized automatically(e.g. boolean for AR 4.2, array for AR 6.1, etc.). That makes possible deserialization of most returning columns. This feature would be nice to have in the context of raw SQL support in returning option, but I see no way how to implement that with back support for projects, which rely on `result.results`. If losing back compatibility is acceptable, I can also add that extra deserialization layer.

jkowens:
Awesome, thanks 🎉

PR #778, opened 2022-08-27
Support on_duplicate_key_update for recursive imports

ramblex:
Hello, this PR adds support for specifying separate `on_duplicate_key_update` options for associations when doing a recursive import. I couldn't find a way of doing that with the existing code - the closest seemed to be passing in :all but that didn't quite do what I needed. Apologies if I've missed if this is already possible..

The PR adds a new `recursive_on_duplicate_key_update` option to `import` where `on_duplicate_key_update` options can be specified for each of the associations. The new option takes a Hash as its value where the keys are the table names and the values are the options normally passed to on_duplicate_key_update.

In the example below  `[:id]` is used as the on_duplicate_key_update option when importing topics and `{ conflict_target: [:id], columns: [:author_name] }` is used as the on_duplicate_key_update option when importing books.

```
Topic.import new_topics,
              recursive: true,
              on_duplicate_key_update: [:id],
              recursive_on_duplicate_key_update: {
                books: { conflict_target: [:id], columns: [:author_name] }
              }
```

Many thanks for this fantastic gem..!


saqlainriaz:
hey @ramblex , thanks for adding this functionality to gem, but can you plz tell me if your code has capability to handle n number of associations with nested child also? **like book has many topics and topic has many sub_topics**

saqlainriaz:
@ramblex i think if you make it recursive then it may fix this problem

railsfactory-dineshr:
I am in need of this feature.
For me, it's just one level of nesting.

ramblex:
Hi @saqlainriaz, recursive_on_duplicate_key_update supports specifying which columns to update for nested associations using something like the following:

```
Topic.import new_topics,
              recursive: true,
              on_duplicate_key_update: [:id],
              recursive_on_duplicate_key_update: {
                books: { conflict_target: [:id], columns: [:author_name] },
                chapters: { conflict_target: [:id], columns: [:title] }
              }
```

I've added in a test to show it working. One possible issue, however, is that the leaf objects (chapters in the example) are only imported if the intermediate objects (books in the example) have been changed, but that seems to be a more general issue - `on_duplicate_key_update: :all` has the same behaviour.

john-h-k:
This would also greatly benefit us

marciondg:
+1 
It would be nice! Is there any update? @zdennis @jkowens

mculp:
this would help us too

jkowens:
Released in v1.7.0!

PR #781, opened 2022-09-15
Rebased upstream

AHaymond:
Adding support for redshift into the base adapater


AHaymond:
Adding support for Redshift

jkowens:
Hi thanks for your contribution. I would consider merging this PR, but it looks like more files than necessary are changed, and the commit history is a little off. Can you get this down to a single commit?

PR #792, opened 2023-02-12
`exec rubocop --auto-gen-config`

smasato:
Update outdated `.rubocop_todo.yml`


mathieujobin:
would probably better to use `--exclude-limit=200` so those with many offenses don't get the rule automatically disabled.

also, enabling rubocop_challenger could help review the rule settings we so desire and auto fix trivial offenses bit by bit....

you can use https://github.com/Apipie/apipie-rails/blob/master/.github/workflows/rubocop-challenger.yml for an example

also, rubocop should be upgraded.

jkowens:
Hey thanks for this nudge. I went ahead and upgraded rubocop and updated the rubocop_todo.yml in #794.

Issue #771, opened 2022-05-27
About track_validation_failures and import methods


Aug-zhong:
Hello
I'm having a problem with this: `track_validation_failures: true`
when i import method take `an array of hashes`, the failed_instances will be `an array of arrays`, with each internal array of the form `[:index_in_dataset, :object_with_errors]` 
when i import method take `an array of models`, the failed_instances will be `an array` with each internal array of the form `[:object_with_errors]`, which loses the `:index_in_dataset`, Is this a bug?

wkirby:
This is a bug, and the culprit is here: https://github.com/zdennis/activerecord-import/blob/v1.4.1/lib/activerecord-import/import.rb#L700

If you are calling import with a set of ActiveRecord Models, the `track_validation_failures: true` setting is never checked.

jkowens:
@wkirby good catch, can you submit a PR?

thimmaiah:
Which version is this fix in? Im running into this in both head version and 1.4.1 of this gem. Thanks

jkowens:
I just released the update for this in v1.5.0.

thimmaiah:
jkowens, this is still not returning the results. Im updated to activerecord-import (1.5.0), yet I get the exact same results as in my post above. Any recommendations ?

PR #798, opened 2023-04-14
add index when models exists

OtaYohihiro:
I confront the same issue as #771 . So, fix it.


jkowens:
Thank you!

jkowens:
Finally released in v1.5.0 🎉

Issue #795, opened 2023-03-07
Allow import of ignored columns - Suggested Feature


kgilje:
I don't know how common this would be for others, but I have models where I obfuscate personal information like username columns from the website interface.  To do this I use the ignore_columns function in the model.  However, inside the import with no validation, the columns are restricted to what you get by model_name.class_hash.  I wrote a work around that reproduced portions of that import code that compared columns instead to ActiveRecord::Base.connection.schema_cache.columns_hash(model_name.table_name) which picks up the the ignored columns and all their attributes in addition to the "normal" columns. 

Where my import breaks with the current code using ignored columns in the input csv.
https://github.com/zdennis/activerecord-import/blob/748309d1ae03fc1fe56d58e54c634a003408c706/lib/activerecord-import/import.rb#L795

Not sure of the complication level of a change like this, but it could be useful to others.

jkowens:
Resolved by #799

Issue #728, opened 2021-03-22
Support for alias_attributes


OskarEichler:
Hey Zach,

First of all - amazing job with this gem! Our database has much fever anxiety attacks thanks to your work!

Secondly: When trying to perform a .import on a model with an alias_attribute, unfortunately we're getting an error message that the column does not exist. 

Would it be possible to add support for that? Thanks so much!

All the best,
Oskar

P.S. Example:

Model - YoutubeArtistRecord
Column - :source_artist_id
alias_attribute - :source_entity_id, :source_artist_id

When performing import with :source_artist_id everything works, but when using :source_entity_id we get an error.

OskarEichler:
I think the implementation for this would be pretty quick, by doing the following:

```
alias_attributes = self.attribute_aliases.symbolize_keys.transform_values!(&:to_sym)
columns.map { |column| alias_attributes[column] || column}
```

jkowens:
Hey @OskarEichler, sorry I just haven't had time to work on this project lately. If you can provide a PR that would be fantastic 👍

jkowens:
Resolved by #799

PR #799, opened 2023-05-10
Introduce support for alias attributes

leonidkroka:
Probably it's related to issues:
https://github.com/zdennis/activerecord-import/issues/728:
```
class MyModel < ActiveRecord::Base
  alias_attribute :foo, :baz
end

MyModel.import(%i[foo bar], %w[value_foo, value_baz], validate: false, on_duplicate_key_update: { conflict_target: %i[bar], columns: %i[foo] })
```

https://github.com/zdennis/activerecord-import/issues/795:
```
class MyModel < ActiveRecord::Base
  self.ignored_columns = [:bar]
end

MyModel.import(%i[foo bar], %w[value_foo, value_baz], validate: false, on_duplicate_key_ignore: true)
MyModel.import(%i[foo bar], %w[value_foo, value_baz], validate: false, on_duplicate_key_update: { conflict_target: %i[foo], columns: %i[bar] })
```

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -574,7 +574,7 @@ class ActiveRecord::Base
 
         if models.first.id.nil?
           Array(primary_key).each do |c|
-            if column_names.include?(c) && columns_hash[c].type == :uuid
+            if column_names.include?(c) && schema_columns_hash[c].type == :uuid
               column_names.delete(c)
             end
           end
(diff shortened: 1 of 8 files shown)


jkowens:
Awesome, thank you for this update! 🎉

jkowens:
@leonidkroka I missed that that 4.2 specs had a failure. I'm ok with dropping support for AR 4.2 if necessary. Should we go that route or can that be resolved easily?

leonidkroka:
@jkowens AR 4.2 has quite a different approach to building `columns_hash`
It's possible to go with if-else condition:
```
if respond_to?(:ignored_columns) && ignored_columns.any?
  connection.schema_cache.columns_hash(table_name)
else
  columns_hash
end
```
Have just opened the respective PR https://github.com/zdennis/activerecord-import/pull/800 (will rerun tests tomorrow)

Yeap, that doesn't look like a cool solution, so it's up to you to use that fix or drop support for AR 4.2

jkowens:
Finally released in v1.5.0 🎉

lib/activerecord-import/import.rb
```diff
@@ -961,6 +964,11 @@ def import_associations(models, options)
       end
     end
 
+    def schema_columns_hash
+      load_schema unless schema_loaded?
+      @schema_columns_hash ||= connection.schema_cache.columns_hash(table_name)
+    end
```
koshigoe, line 970: @leonidkroka @jkowens 

I think the `@schema_columns_hash` should be cleared by `.reset_column_information`.

👉 https://github.com/zdennis/activerecord-import/issues/809

PR #801, opened 2023-05-19
Support for multi-byte column names when splitting queries

TakuyaKurimoto:
Hello !
Query splitting does not work well when column names contain multibyte characters.
Here is an example of a table schema that causes problems.
```
create_table :topics, force: :cascade do |t|
    t.string 'あいう'  
end
```


jkowens:
Nice, thank you!

jkowens:
Finally released in v1.5.0 🎉

PR #802, opened 2023-05-23
feat: Omit Columns from SQL statements (e.g uuid columns)

Amnesthesia:
I've added two options for omitting columns from the SQL statements, to avoid having to modify class level `self.ignored_columns` since we've had issues with this affecting other code that runs at the same time. These options work as follows:

Omit the guid column from SQL statement, allowing it to generate the uuid and not be forced to NULL:
`Model.import(values, omit_columns: [:guid])` 

Allow per-model decisions, e.g for recursive imports:
`Model.import(values, omit_columns: -> (model, column_name) { [:guid] if model == Model })`

Use a hash instead of a proc:
`Model.import(values, omit_columns: { Model => [:guid] })`

The second option is `:omit_columns_with_default_functions` boolean, to automatically find columns that have a default function declared in the schema, and omit them by default.


Amnesthesia:
Can this be merged or is this already implemented upstream? We're currently pointing our Gemfile to this branch and can't update until uuid primary keys are supported

jacob-carlborg-apoex:
@Amnesthesia UUID primary keys already work, if they're following the standard naming convention, i.e. `id`. I'm using that on several tables in one of my applications. But for other columns it doesn't work.

jacob-carlborg-apoex:
@Amnesthesia BTW, it's already possible to specify which columns should be imported. Pass an array of symbols as the first argument to `import`. Second code example: https://github.com/zdennis/activerecord-import?tab=readme-ov-file#activerecord-models.

Amnesthesia:
@jacob-carlborg-apoex This is for UUID columns that are not primary keys :)

Issue #787, opened 2022-11-30
Upcoming breakage with next Rails release


gravitystorm:
I was working on something completely unrelated, and came across this problem with using the latest activerecord-import version running against the current rails `main` branch.

https://github.com/zdennis/activerecord-import/blob/ef08aca875b956c74f0432e78b0fa1559e189eb9/lib/activerecord-import/import.rb#L88

The signature of the Rails internal `CallbackChain.compile` method has recently (September 2022) changed, and now requires a `type` parameter (although that type can be nil, it's not optional).

```diff
-        def compile
+        def compile(type)
```

https://github.com/rails/rails/blob/a51e9676e611f578bdfd10fcf6891d9e2d762a23/activesupport/lib/active_support/callbacks.rb#L651-L667

It was changed in:
* https://github.com/rails/rails/commit/b985dc2b1fddccb82cb34386c94475f3c00d0e4b
* https://github.com/rails/rails/pull/45952

The error message I got in my app was "ArgumentError: wrong number of arguments (given 0, expected 1)".

I thought I'd open this issue here to give you a heads-up! I don't have any particular urgency or anything with this, so no worries if it's low priority (or ignored) until the next official Rails release is made.

Morozzzko:
Just ran into the issue. Running 7.1.0 Alpha from the `main` branch

Had to downgrade Rails to unblock myself. Can't wait to see the issue fixed

mathieujobin:
I confirm the issue.

I made this branch simply sending nil.

https://github.com/zdennis/activerecord-import/compare/master...mathieujobin:activerecord-import:rails-7.1

beeleethebee:
Hi guys ! Any update on this ? I'm facing the same issue. Is it possible to merge the fix ?

gjtorikian:
👋 I began to implement what @jkowens was asking for:

```diff
diff --git a/lib/activerecord-import/import.rb b/lib/activerecord-import/import.rb
index 11eeab7..d761a19 100644
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -85,7 +85,12 @@ module ActiveRecord::Import # :nodoc:
 
         model.run_callbacks(:validation) do
           if defined?(ActiveSupport::Callbacks::Filters::Environment) # ActiveRecord >= 4.1
-            runner = @validate_callbacks.compile
+            runner = if @validate_callbacks.method(:compile).arity == 0
+                       @validate_callbacks.compile
[...]

PR #807, opened 2023-08-11
Send nil to ActiveSupport#compile which now requires a parameter

gucki:
Fixes https://github.com/zdennis/activerecord-import/issues/787


gucki:
@jkowens You are welcome! Would you mind releasing a new version with the fix? :)

PR #813, opened 2023-09-03
Feature:  Add Dockerfile and compose file

zmariscal:
# Purpose
Make it easier for people to develop using Docker.

# Implementation
Add a Dockerfile and docker-compose.yml to help developers be able to pull an image and run the application locally. You can read the README for how to do this. I went for the most minimal change to get mysql working first.

# Potential future work
The next step would be to add `postgresql` as a service so someone can run the tests against that database service.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+# Use the official Ruby 3.2 image as a base image
+ARG RUBY_VERSION=3.2
+FROM ruby:${RUBY_VERSION}-bullseye
+
+# Set the working directory
+WORKDIR /usr/src/app
+
+# Install system packages
+RUN apt-get update -qq && \
+    apt-get install -y default-mysql-client vim && \
+    apt-get clean
+
+# Set environment variables
+ENV AR_VERSION=7.0
+
+# Copy all files
+COPY . .
+
+# Move sample database.yml and install gems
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+    build:
+      context: .
+    environment:
+      DB_HOST: mysql
+      AR_VERSION: 7.0
+    volumes:
+      - .:/usr/src/app
+    depends_on:
+      - mysql
+    command: tail -f /dev/null
```
zmariscal, line 22: There may be a better command to use here. It's just the simplest thing I could think of to ensure the container doesn't exit.

Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Use the official Ruby 3.2 image as a base image
+ARG RUBY_VERSION=3.2
+FROM ruby:${RUBY_VERSION}-bullseye
+
+# Set the working directory
+WORKDIR /usr/src/app
+
+# Install system packages
+RUN apt-get update -qq && \
+    apt-get install -y default-mysql-client vim && \
```
zmariscal, line 10: Adding `vim` isn't necessary, but it sure makes things easier to debug in container

jkowens:
Thanks @zmariscal this is a great idea!

PR #817, opened 2023-09-07
Import breaks when two column backed attributes are aliased together

skunkworker:
When a record has two column backed attributes that are aliased together, import breaks across the board because both column names are provided. I briefly looked to see if this was a code smell in ActiveRecord but have not been able to find anything definitive either way.

Take the following:

```ruby
create_table :topics, force: :cascade do |t|
  t.string :author_name
  t.string :author_full_name
end

class Topic < ActiveRecord::Base
  alias_attribute :author_full_name, :author_name
end
```

When the record is trying to be imported. `author_name` will be specified twice in the import string.

### DIFF
--- a/test/import_test.rb
+++ b/test/import_test.rb
@@ -991,4 +991,16 @@ describe "#import" do
       end
     end
   end
+
+  describe "aliased attributes" do
+    it "two column based attributes can be imported" do
+      topic = FactoryBot.create :topic
+      new_topic = Topic.new(topic.attributes.except("id"))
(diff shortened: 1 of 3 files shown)


jkowens:
Aha! Ok, thank you for getting to the root issue.

jkowens:
@leonidkroka here is a some more info about a failing scenario with `alias_attribute`. Any ideas how to make sure column names in the generated sql are unique in this scenario?

PR #820, opened 2023-10-26
Bump `Checkout` action to V4

y-yagi:
The `Checkout` action V2 uses old Node.js. It causes a lot of warnings and bump version to fix those.

Ref: "Annotations" section in https://github.com/zdennis/activerecord-import/actions/runs/6556382678


jkowens:
Thanks!

PR #822, opened 2023-11-22
Use ActiveRecord::Base.locking_enabled? to update the lock field

dombesz:
### Problem

`activerecord-import` is ignoring the [`ActiveRecord::Base.lock_optimistically = false`](https://api.rubyonrails.org/classes/ActiveRecord/Locking/Optimistic.html) flag and it still updates the lock field.

### Solution

Use the [`locking_enabled?`](https://api.rubyonrails.org/classes/ActiveRecord/Locking/Optimistic/ClassMethods.html#method-i-locking_enabled-3F) method provided by activerecord to decide whether the lock field should be updated.


jkowens:
Makes sense, thanks!

jkowens:
Released in v1.6.0 👍

PR #825, opened 2024-02-03
[Feature]  Add trilogy adapter support

zmariscal:
# Purpose

Bring support for the [Trilogy adapter](https://github.blog/2022-08-25-introducing-trilogy-a-new-database-adapter-for-ruby-on-rails/)to this project. I work on a Rails project and would like to switch to this new adapter. When running our CI I discovered this gem didn’t have support. Which then pointed to a child dependency that needed that support. I added that [here](https://github.com/composite-primary-keys/composite_primary_keys/pull/607).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -95,6 +95,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
+          rubygems: latest
       - name: Set up databases
         run: |
           sudo /etc/init.d/mysql start
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -16,6 +16,20 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
+      mysql:
+        image: mysql:5.7
+        ports:
+          - 3306:3306
+        env:
+          MYSQL_ROOT_PASSWORD: root
+          MYSQL_USER: github
+          MYSQL_PASSWORD: github
+          MYSQL_DATABASE: activerecord_import_test
+        options: >-
+          --health-cmd "mysqladmin ping -h localhost"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
     strategy:
       fail-fast: false
       matrix:
@@ -110,6 +124,9 @@ jobs:
         run: |
           bundle exec rake test:spatialite
           bundle exec rake test:sqlite3
+      - name: Run trilogy tests
+        if: ${{ matrix.env.AR_VERSION >= '7.0' }}
+        run: bundle exec rake test:trilogy
   lint:
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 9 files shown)


.github/workflows/test.yaml
```diff
@@ -81,6 +95,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
+          rubygems: latest
```
zmariscal, line 98: I added [this](https://github.com/ruby/setup-ruby/blob/22fdc77bf4148f810455b226c90fb81b5cbc00a7/action.yml#L13) because [this error](https://github.com/zdennis/activerecord-import/actions/runs/7768197459/job/21185855381#step:6:10) was raised. This should ensure this type of error doesn't happen again.

.github/workflows/test.yaml
```diff
@@ -110,6 +125,9 @@ jobs:
         run: |
           bundle exec rake test:spatialite
           bundle exec rake test:sqlite3
+      - name: Run trilogy tests
+        if: ${{ matrix.env.AR_VERSION >= '7.0' && matrix.ruby != 'jruby' }}
```
zmariscal, line 129: I'm not entirely sure where the trilogy gem stands with jruby, but I couldn't figure out a way to get it to successfully run.

Gemfile
```diff
@@ -26,6 +26,10 @@ platforms :ruby do
   gem "sqlite3",                "~> #{sqlite3_version}"
   # seamless_database_pool requires Ruby ~> 2.0
   gem "seamless_database_pool", "~> 1.0.20" if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('3.0.0')
+  gem "trilogy" if version >= 6.0
+  if version >= 6.0 && version <= 7.0
+    gem "activerecord-trilogy-adapter"
+  end
```
zmariscal, line 32: Regarding the integration of Trilogy, theoretically, it should seamlessly function with any framework equipped with the `active record-trilogy-adapter`. Notably, Rails 7.1 conveniently incorporates this adapter by default. You can explore the prerequisites for this gem [here](https://github.com/trilogy-libraries/activerecord-trilogy-adapter?tab=readme-ov-file#requirements). However, a noteworthy discovery I made is the importance of the `composite_primary_keys` version. Compatibility issues arise with versions preceding AR 7 due to limitations intertwined with the ActiveRecord version.

Gemfile
```diff
@@ -37,7 +41,11 @@ platforms :jruby do
 end
 
 # Support libs
-gem "factory_bot"
+if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.0.0")
+  gem "factory_bot"
+else
+  gem "factory_bot", "~> 5", "< 6.4.5"
+end
```
zmariscal, line 48: This is required due to the newest version of [factory_bot](https://github.com/thoughtbot/factory_bot/issues/1614) requiring Ruby 3.0.

jkowens:
Thanks @zmariscal! This is awesome 🙌

PR #831, opened 2024-03-01
CI against Ruby 3.3

y-yagi:
This PR adds Ruby 3.3 to the CI matrix to ensure `activerecord-import` work with Ruby 3.3


y-yagi:
It looks like JRuby 9.4.6.0 + `activerecord-jdbcmysql-adapter` raises an error. I'm doubting this is the same issue with https://github.com/jruby/jruby/issues/8116.

jkowens:
Thank you!

PR #834, opened 2024-03-15
Fix jruby spec runs

zmariscal:
# Purpose
With the new release of this gem we updated the version of jruby https://github.com/zdennis/activerecord-import/commit/0b22f67d5d0f69f86244de8caee658ba80b56430. 

# Fix
We were currently matching on `jruby`, but let's make sure we never run against `jruby` if the  matrix ruby contains `jruby`.


jkowens:
Awesome, thank you! For some reason `jruby` stable when running the mysql tests. I pinned it to `jruby-9.4.5.0` but that should probably be addressed at some point.

Issue #830, opened 2024-02-23
Error with composite PK


fragkakis:
I have trouble importing data to a table with a composite PK. Here is the scenario that can reproduce this in Postgres:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", "~> 7.1.3"
  gem 'activerecord-import'
  gem "pg"
end

require "active_record"
require "minitest/autorun"
require "logger"

db_properties = {
  adapter: "postgresql",
  host: "localhost",
  username: "postgres",
  password: "postgres",
  database: "books"
}

ActiveRecord::Base.establish_connection(db_properties)
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :authors, force: true do |t|
    t.string :name
  end

  execute <<-SQL
    DROP SEQUENCE IF EXISTS book_id_seq;    
    CREATE SEQUENCE book_id_seq
        AS integer
        START WITH 1
        INCREMENT BY 1
        NO MINVALUE
        NO MAXVALUE
        CACHE 1;

    DROP TABLE IF EXISTS books; 
    CREATE TABLE books (
        id bigint DEFAULT nextval('book_id_seq'::regclass) NOT NULL,
        title character varying,
        author_id bigint
     );

    ALTER TABLE ONLY books ADD CONSTRAINT fk_rails_040a418131 FOREIGN KEY (author_id) REFERENCES authors(id);
  SQL
end

class Author < ActiveRecord::Base
  has_many :books, query_constraints: [:id, :author_id], inverse_of: :author
end

class Book < ActiveRecord::Base
  self.primary_key = %i[id author_id]
  belongs_to :author
end

author = Author.create!(name: "Foo Barson")

books = []
2.times do |i|
  books << Book.new(author_id: author.id, title: "book #{i}")
end
Book.import books
```

The error I am getting is:

```ruby
/Users/markosfragkakis/.rvm/gems/ruby-3.2.2/gems/activerecord-7.1.3.2/lib/active_record/connection_adapters/abstract/quoting.rb:29:in `quote': can't quote Array (TypeError)

        else raise TypeError, "can't quote #{value.class.name}"
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #837, opened 2024-04-22
Rails 71 composite pks

fragkakis:
Adding support for Rails 7.1 composite primary keys (issue #830 )

For Rails 7.1 I am making the assumption that the [composite-primary-keys](https://github.com/composite-primary-keys/composite_primary_keys) gem is not going to be used, since Rails 7.1 has out-of-the-box support for composite PKs. I have added some inline comments in the PR for the reviewers.

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -941,7 +941,7 @@ class ActiveRecord::Base
           association = association.target
           next if association.blank? || model.public_send(column_name).present?
 
-          association_primary_key = Array(association_reflection.association_primary_key)[column_index]
+          association_primary_key = Array(association_reflection.association_primary_key.tr("[]:", "").split(", "))[column_index]
           model.public_send("#{column_name}=", association.send(association_primary_key))
         end
       end
@@ -996,7 +996,10 @@ class ActiveRecord::Base
 
         changed_objects = association.select { |a| a.new_record? || a.changed? }
         changed_objects.each do |child|
-          child.public_send("#{association_reflection.foreign_key}=", model.id)
+          Array(association_reflection.inverse_of&.foreign_key || association_reflection.foreign_key).each_with_index do |column, index|
+            child.public_send("#{column}=", Array(model.id)[index])
+          end
+
           # For polymorphic associations
           association_name = if model.class.respond_to?(:polymorphic_name)
             model.class.polymorphic_name
--- a/.gitignore
+++ b/.gitignore
@@ -13,12 +13,16 @@ tmtags
 ## VIM
 *.swp
 
+## Idea
+.idea
+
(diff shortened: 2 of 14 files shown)


lib/activerecord-import/import.rb
```diff
@@ -941,7 +941,7 @@ def load_association_ids(model)
           association = association.target
           next if association.blank? || model.public_send(column_name).present?
 
-          association_primary_key = Array(association_reflection.association_primary_key)[column_index]
+          association_primary_key = Array(association_reflection.association_primary_key.tr("[]:", "").split(", "))[column_index]
```
fragkakis, line 944: Without a composite PK, the association_reflection.association_primary_key has a value of the format "id", whereas with a composite PK in ActiveRecord 7.1 the value is of the format "[:id, :account_id]" (again a String). The .tr loses the [, :, ] characters, so that the .split(", ") will give us an array of the two colums: ["id", "account_id"]

lib/activerecord-import/import.rb
```diff
@@ -996,7 +996,10 @@ def find_associated_objects_for_import(associated_objects_by_class, model)
 
         changed_objects = association.select { |a| a.new_record? || a.changed? }
         changed_objects.each do |child|
-          child.public_send("#{association_reflection.foreign_key}=", model.id)
+          Array(association_reflection.inverse_of&.foreign_key || association_reflection.foreign_key).each_with_index do |column, index|
+            child.public_send("#{column}=", Array(model.id)[index])
+          end
+
```
fragkakis, line 1002: Another note for the reviewers.

Let's now consider we want to recursively import the model CompositeBook and its children CompositeChapter.

[Here](https://github.com/fragkakis/activerecord-import/blob/rails_71_composite_pks/test/models/composite_book.rb#L12-L13) is the has_many relation on CompositeBook, mentioning the names of the columns id and author_id. These are the PK columns in the composite_books table.

As we are setting the ids in the children, we need the columns as they are called in the children table (composite_chapters), which is [here](https://github.com/fragkakis/activerecord-import/blob/rails_71_composite_pks/test/models/composite_chapter.rb#L5-L6). So we need the inverse relation (the belongs_to in CompositeChapter), which has the names composite_book_id and author_id.

In order not to break things, I kept the original association_reflection.foreign_key as a fallback.

jkowens: Looks good 👍 If an association doesn't explicitly state the `inverse_of` does AR infer that relationship so that this still works?

jkowens (approved):
This looks good to me! Thanks @fragkakis 🙌  It doesn't sound like there will be any CPK gem releases for AR 7.1, so I think it's safe to end support with AR 7.0.

jkowens:
Released in v1.7.0!

PR #850, opened 2024-09-06
Fix Ruby 3.3.5 warning - ostruct will no longer be part of the default gems

shaicoleman:
Fixes the warning with Ruby 3.3.5:

warning: .../ruby/3.3.5/lib/ruby/3.3.0/ostruct.rb was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0. You can add ostruct to your Gemfile or gemspec to silence this warning.

https://github.com/ruby/ruby/blob/v3_3_5/lib/bundled_gems.rb#L29


shaicoleman:
Alterntive PR: #853

jkowens:
Closed in favor of #853 👍

PR #852, opened 2024-09-17
Fix #843: recursive_on_duplicate_key_update doesn't work with non-standard association name

### LINKED ISSUE #843: recursive_on_duplicate_key_update doesn't work with non-standard association name
If I have a class structure like this:

```ruby
class Foo < ActiveRecord::Base
  has_many :items, class_name: "FooItem"
end

class FooItem < ActiveRecord::Base
end
```

And try to import using `recursive_on_duplicate_key_update`, like this:

```ruby
Foo.import(records,
  on_duplicate_key_update: {columns: :all},
  recursive_on_duplicate_key_update: {items: {columns: :all}})
```

Then activerecord-import won't do an upsert for the association, only for the main records.

The issue is that `options[:recursive_on_duplicate_key_update][table_name]` returns `nil` here:

https://github.com/zdennis/activerecord-import/blob/fca8b823ae695b03714837cc6603f51525c60505/lib/activerecord-import/import.rb#L865

The code is looking for a key matching the table name of the association instead of the name of the association itself. The documentation says: "The hash key is the association name".

jacob-carlborg-apoex:


jacob-carlborg-apoex:
Failures are due to the MySQL server not starting.

jkowens:
Looks great, thank you!

PR #853, opened 2024-09-20
Remove OpenStruct

shaicoleman:
Fixes the warning with Ruby 3.3.5:
warning: .../ruby/3.3.5/lib/ruby/3.3.0/ostruct.rb was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0. You can add ostruct to your Gemfile or gemspec to silence this warning.

OpenStruct use officially discouraged
https://docs.ruby-lang.org/en/master/NEWS/NEWS-3_0_0_md.html
https://docs.ruby-lang.org/en/master/OpenStruct.html#class-OpenStruct-label-Caveats

### DIFF
--- a/benchmarks/lib/base.rb
+++ b/benchmarks/lib/base.rb
@@ -16,7 +16,7 @@ class BenchmarkBase
     end
   end
 
-  # Returns an OpenStruct which contains two attritues, +description+ and +tms+ after performing an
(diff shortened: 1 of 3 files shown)


epugh:
Looking for this too!

jkowens:
Looks good, thank you!

PR #871, opened 2025-03-11
Remove hash content from key mismatch error message

### LINKED ISSUE #868: Hash key mismatch error message can cause data leakage
**Description**
When importing an array of hashes, if there is a key mismatch in the provided data, the error message exposes the whole hash revealing potentially sensitive data.

**Example**
```
class BooksController < ApplicationController
  def bulk_create
    books = Book.import(books_params)
    render json: { books: }, status: (books.failed_instances.any? ? :unprocessable_entity : :ok)
  rescue StandardError => e
    render json: { error: e }, status: :bad_request
  end
end
```
Curl request:
```
curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer <JWT_TOKEN>" localhost:3000/books/bulk_create -d '{
  "books": [
    { "title": "book1", "description": "book1" },
    { "titlez": "book2", "description": "book2" }
  ]
}'
```
This request contains a typo (titlez instead of title) in one of the hashes. The resulting error message is as follows:
```
{
  "error": "Hash key mismatch.\n\nWhen importing an array of hashes, all hashes must have the same keys.\nIf you have records that are missing some values, we recommend you either set default values\nfor the missing keys or group these records into batches by key set before importing.\n\nRequired keys: [\"title\", \"description\", \"author_id\"]\nExtra keys: []\nMissing keys: [\"title\"]\n\nHash: {\"description\"=>\"book2\", \"author_id\"=>1}\n"
}
```
In this application the author ID is assigned automatically and we don't want the user to know it for security reasons (though it could be any generated or assigned column data).

**Suggested Fix**
Remove the "Hash" section from the error message, or provide an option to pass column whose values will be filtered.

smasato:
Fix #868
Test / lint (pull_request) is failing.
Fixed in https://github.com/zdennis/activerecord-import/pull/870

### DIFF
--- a/lib/activerecord-import/import.rb
+++ b/lib/activerecord-import/import.rb
@@ -1132,8 +1132,6 @@ When importing an array of hashes with provided columns_names, each hash must co
 
 Required keys: #{required_keys}
 Missing keys: #{missing_keys}
-
-Hash: #{hash}
         EOS
       else
         <<-EOS
@@ -1146,8 +1144,6 @@ for the missing keys or group these records into batches by key set before impor
 Required keys: #{required_keys}
 Extra keys: #{extra_keys}
 Missing keys: #{missing_keys}
-
-Hash: #{hash}
         EOS
       end
     end


jkowens:
Thanks!

PR #876, opened 2025-06-04
Ooops. Actually replace all #connection with #with_connection

nerdrew:
ActiveRecord 8.0 deprecated `#connection`, and encourages users to use `#with_connection` or `#lease_connection`. `#with_connection` feels more reasonable for a library to use because it doesn't impact the overall connection management for the app using the library.

Instead of getting a connection for each usage, I checkout a connection at the "entry point" to the library and use that connection whenever it is needed. This unfortunately requires passing a connection object around a bit...


nerdrew:
Any thoughts on this PR?

jkowens:
Sorry for the delay. I think it looks good, thanks 👍

PR #880, opened 2025-08-27
Fix ActiveRecordProxyAdapters support

mateuscruz:
Hi there! I'm the maintainer of `active_record_proxy_adapters`.

`ActiveRecordProxyAdapters` works on the premise that an incoming SQL string could be either a read or a write and has some routing mechanism in place to decide which database it should send the request too. That uses `Regexp` pattern matching which, in most cases, should perform pretty quickly but in the context of this gem, it might be a bottleneck depending on how long the query string is - since it's meant for bulk imports.

Because we know for a fact that the imports are always going to be writes, it is safe to bypass most of the routing logic by wrapping the calls within a `connected_to` block (introduced in Rails 6). This reduces the pattern matching overhead which would happen for every bulk import and allows the queries to go straight to the writer instance.

I decided to make this change after a bug was reported which was also impacting activerecord-import users: https://github.com/Nasdaq/active_record_proxy_adapters/issues/113

Thanks!

### DIFF
--- a/lib/activerecord-import/active_record/adapters/mysql2_proxy_adapter.rb
+++ b/lib/activerecord-import/active_record/adapters/mysql2_proxy_adapter.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+require "active_record/connection_adapters/mysql2_proxy_adapter"
+require "activerecord-import/adapters/mysql2_proxy_adapter"
+
+ActiveSupport.on_load(:active_record_mysql2proxyadapter) do |klass|
+  klass.include(ActiveRecord::Import::Mysql2ProxyAdapter)
+end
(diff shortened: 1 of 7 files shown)


mateuscruz:
@jkowens tagging for visibility.

Also, if you're good with these changes, I'd be glad to open a follow-up MR to add support for Trilogy and Sqlite3 proxies as well (I see those are missing here).

jkowens:
Hi @mateuscruz, sorry I haven't really had a chance to look at this too closely. I haven't done much with this library in a while so my memory is getting foggy 😅 

Anyway, I'd be happy to add the proxy adapters but I'm not sure I follow what you said about n + 1 inserts. That part does concern me a bit. It sounds like that will happen across the board unless users of the library change their code?

mateuscruz:
@jkowens hi. Just double checking if there's still any interest in having this merged? Thanks.

jkowens:
Great, thanks @mateuscruz 👍

PR #885, opened 2025-09-12
Fix affected_rows data source for ON DUPLICATE KEY UPDATE operations

ehasrouni:
Problem: `affected_rows` always returned `0` for `ON DUPLICATE KEY UPDATE` operations in Trilogy adapter.

Root Cause: In this [PR](https://github.com/zdennis/activerecord-import/pull/882), we had used `raw_connection.affected_rows` which returns stale/incorrect values for ON DUPLICATE KEY UPDATE in Trilogy.

The correct values are available in `result.affected_rows` from the execute result object, so this simply changes the implementation to make use of the result rather than the connection to fetch the `affected_rows`, super simple change.

This PR:
Switch from `raw_connection.affected_rows` to `result.affected_rows` when available
Preserve `nil` for adapters without support
Adds a test to cover for that edge case
Fix batch accumulation to handle nil values properly

### DIFF
--- a/lib/activerecord-import/adapters/mysql_adapter.rb
+++ b/lib/activerecord-import/adapters/mysql_adapter.rb
@@ -12,7 +12,7 @@ module ActiveRecord::Import::MysqlAdapter
   def insert_many( sql, values, options = {}, *args ) # :nodoc:
     # the number of inserts default
     number_of_inserts = 0
-    affected_rows = raw_connection.respond_to?(:affected_rows) ? 0 : nil
+    affected_rows = nil
 
     base_sql, post_sql = case sql
                          when String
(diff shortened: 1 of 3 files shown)


ehasrouni:
@jkowens hey! this should be the last piece of the puzzle (a small bug fix to what we [merged last week](https://github.com/zdennis/activerecord-import/pull/882)). It just adds support for `ON DUPLICATE KEY UPDATE` operations that were returning 0 with the Trilogy adapter by simply using the result's `affected_rows` rather than the `raw_connection`. Business as usual for unsupported adapters.

Thanks a lot! let me know if you have any other concerns

jkowens:
Thanks @ehasrouni 👍

PR #887, opened 2025-09-24
Add support for trilogy and sqlite proxy adapters

mateuscruz:
Hello! As promised, here's the PR for the missing proxy adapters: `Trilogy` and `SQLite3`.


mateuscruz:
@jkowens could you take a look when you have a moment? Thanks.

PR #889, opened 2025-12-18
add proxy adapters to 1.8.0

nlively:
this pr backports @mateuscruz changes onto version 1.8.0 to make this version compatible with the `active_record_proxy_adapters` gem


jkowens:
Thanks for the PR. I’m going to close this because the proxy adapter support already exists on current `master` for mysql2, postgresql, sqlite3, and trilogy.

This also says it backports support to `1.8.0`, but the PR targets `master`, so it does not apply as a release-branch backport. If there’s still a need for a `1.8.x` branch release, we’d need that as a PR against the appropriate maintenance branch rather than `master`.

PR #892, opened 2026-01-08
Move column_names assignment inside returning_selections

mirelon:
When running the import, I was getting error from deep inside this gem:
```
undefined method 'column_names' for nil (NoMethodError)
```

Something is probably wrong on more layers, but the fix on this place is **simple** and **safe**, because `column_names` are not needed if there are no `selections`.

### DIFF
--- a/lib/activerecord-import/adapters/postgresql_adapter.rb
+++ b/lib/activerecord-import/adapters/postgresql_adapter.rb
@@ -98,12 +98,12 @@ module ActiveRecord::Import::PostgreSQLAdapter
 
   def returning_selections(options)
     selections = []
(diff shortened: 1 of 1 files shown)


jkowens:
I agree with avoiding the `column_names` lookup when there are no selections, but I don’t think we should move the assignment inside the `map`. That repeats the same lookup for every selection and makes the method harder to read.

Could we instead build selections first, then handle missing model metadata explicitly? For example:

```ruby
column_names = Array(options[:model]&.column_names)

selections += Array(options[:primary_key]) if options[:primary_key].present?
selections += Array(options[:returning]) if options[:returning].present?
return selections if column_names.empty?

PR #898, opened 2026-07-27
Document #886's fix in CHANGELOG (Unreleased)

leonardow-unep-wcmc:
## Summary

While debugging why `activerecord-import`'s adapter-specific support (e.g. `on_duplicate_key_update`) wasn't loading for a secondary MySQL connection set up via `connects_to` in a Rails 8 multi-database app, I traced it to exactly what #886 already fixed - moving the lazy adapter-loading hook from `ActiveRecord::Base.establish_connection` (only ever called for the primary connection at boot) to `ActiveRecord::ConnectionAdapters::ConnectionHandler#establish_connection` (fires for every connection, regardless of how it was established).

That fix is merged to `master` but isn't mentioned in the `Unreleased` section of CHANGELOG.md yet - this PR just adds that entry, so it's visible ahead of the next release.

No code changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jkowens:
Thank you 👍

PR #900, opened 2026-08-07
Fix NoMethodError on Rails main by no longer requiring ActiveModel's private validation_context= writer

yu-yaba:
## Problem

Rails removed the private `ActiveModel::Validations#validation_context=` writer in
rails/rails@0f9d1270834a6407a59637650bf910d8ae826169 (on main as of
rails/rails@bcce5106f8b3d074a91d6efaa4a34b7e4a087f3c, version 8.2.0.alpha), in favor of a
validation context object reached through the private `#context_for_validation` reader.

`ActiveRecord::Import::Validator#valid_model?` calls that writer twice — once to set the
import's validation context and once to restore the caller's context in `ensure`. Against
Rails main, every validated import therefore blows up with:

    NoMethodError: undefined method 'validation_context=' for an instance of <Model>

This affects activerecord-import 2.2.0 and current master. Running the sqlite3 suite against
Rails main before this change gives 3 failures and 114 errors.

## Fix

Add a small private `Validator#set_validation_context(model, context)` helper that uses
`validation_context=` when the writer is present and falls back to
`context_for_validation.context=` otherwise, and use it in both places. The branch is chosen
by method presence rather than by ActiveRecord version, so a single code path serves
ActiveRecord 5.2 through main. Import validation is not disabled and no Rails monkey patch
is introduced.

## Tests

`test/import_test.rb` gains a "with respect to the model's validation context" context
covering valid imports, invalid imports, `validate_with_context`, restoration of the previous
validation context, and restoration when validation raises.

Because every released ActiveModel still defines `validation_context=`, the new fallback
branch would be dead code on the entire existing matrix. `test/support/rails_main_validation_context_model.rb`
adds a small `ActiveModel::Validations` model that removes the writer and implements
`context_for_validation` the way Rails main does, so the fallback is genuinely exercised on
every supported Rails version. Reverting the library change makes 7 of the new tests error.

## CI

Adds `gemfiles/8.2.gemfile`, which tracks the `rails/rails` main branch, plus a matrix entry
for `AR_VERSION=8.2` on Ruby 3.4 (Rails main requires Ruby >= 3.3.1). The job is marked
`experimental` and runs under `continue-on-error`, so upstream churn cannot turn the matrix
red, and it skips `bundler-cache` — a cached lockfile would pin the git source to a stale
revision instead of resolving the current branch head.

## Results

* Rails main (8.2.0.alpha): 224 runs, 560 assertions, 0 failures, 0 errors (was 3 failures,
  114 errors)
* ActiveRecord 8.1, 8.0, 7.2, 7.1, 7.0, 6.1 — sqlite3, sqlite3_proxy, spatialite: all green
* ActiveRecord 8.0 and 7.0 — mysql2, postgresql: all green
* RuboCop: 145 files inspected, no offenses detected

## Note on private APIs

`context_for_validation` is private on Rails main, so this trades one private-API dependency
for another. There is no public way to swap a model's validation context without also running
validations (`valid?(context)`), which `valid_model?` cannot use since it drives the validate
callback chain itself. A public ActiveModel API for scoping a validation context would let
this shim go away entirely.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -51,6 +51,12 @@ jobs:
           - AR_VERSION: '6.1'
             RUBYOPT: --enable-frozen-string-literal
         include:
+          # Rails main (8.2.0.alpha). Allowed to fail so that in-flight upstream changes do
+          # not turn the whole matrix red; see continue-on-error below.
+          - ruby: 3.4
+            experimental: true
+            env:
+              AR_VERSION: '8.2'
           - ruby: 3.3
             env:
               AR_VERSION: '8.1'
@@ -118,6 +124,7 @@ jobs:
             env:
               AR_VERSION: 5.2
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental || false }}
     env:
       AR_VERSION: ${{ matrix.env.AR_VERSION }}
       DB_DATABASE: activerecord_import_test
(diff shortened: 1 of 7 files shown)


jkowens:
Looks good, thank you 👍

